Issue #3, opened 2015-11-12
build: removing guice dependency


mgenov:
If Guice depenedencies are removed from library, it will make it more flexible and usable from code which is not using Guice. 

Also the usage will be simplified somehow, cause API design would be changed too.

PR #5, opened 2016-02-22
oauth2: initial step for removing of guice dependencies

mgenov:
Currently uppper level parts of the code are using Guice for the frontend part. This change is moving one direction ahead by introducing a single servlet class which will perform object wiring and will dispatch every OAuth2 request to the internal backbone.

Issuing of new tokens and refreshing of existing tokens is now performed from the newly created OAuth2Servlet.


mgenov:
Related to #3

PR #8, opened 2016-03-24
oauth2: add logging-in support using cookies

mgenov:
The existing logging-in support wasn't working till know with the manual wiring of objects and OAuth2Servlet which is now fixed by improving the design of existing interfaces.


tsony-tsonev:
LGTM

PR #18, opened 2016-09-30
oauth2: revoke token

### LINKED ISSUE #16: oauth2: revoke access token of an user
When client application wants to log-out an user it has to revoke token in the oauth2 provider. 

The complete RFC description is: 
https://tools.ietf.org/html/rfc7009#section-2.1

mgenov:
Added logic for revoking existing token from the client.

This change completely implements points 2.1 and 2.2 from https://tools.ietf.org/html/rfc7009

Fixes #16


oauth2-server/src/main/java/com/clouway/oauth2/token/Tokens.java
```diff
@@ -31,9 +30,18 @@
...
+   */
+  Token issueToken(Client client, String identityId, DateTime when);
+
+  /**
+   * Revokes token from repository.
+   *
+   * @param token the token which to be revoked
    */
-  Token issueToken(String identityId, DateTime when);
+  void revokeToken(String token);
```
mlesikov, line 46: I can't fin the implemntation of this method

mgenov: Tokens interface is implemented in the applications which are consuming oauth2-server library.

mlesikov: Ok

mlesikov (approved):

oauth2-server/src/main/java/com/clouway/oauth2/token/Tokens.java
mgenov: The InMemory implementation in the example app now contains such implementation.

PR #22, opened 2016-10-06
oauth2: recognise different identities by GrantType

### LINKED ISSUE #21: oauth2: userInfo endpoint should return user info for service accounts
When userInfo is requested with token issued for service account it should return proper information for it. Currently interface is design as follow: 

``` java
public interface IdentityFinder {
  Optional<Identity> findIdentity(String identityId, DateTime instantTime);
}
```

and the problem is that implementations wouldn't be able easily to distinguish between identities of service accounts and the resource owners.

mgenov:
Token now contains and grant type which could be used by the applications to separate Service Accounts from Resource Owners.

Fixes #21

### DIFF
--- a/oauth2-example-app/src/main/java/com/clouway/oauth2/exampleapp/storage/InMemoryUserRepository.java
+++ b/oauth2-example-app/src/main/java/com/clouway/oauth2/exampleapp/storage/InMemoryUserRepository.java
@@ -4,6 +4,7 @@ import com.clouway.friendlyserve.Request;
 import com.clouway.oauth2.DateTime;
 import com.clouway.oauth2.Identity;
 import com.clouway.oauth2.exampleapp.UserRepository;
+import com.clouway.oauth2.token.GrantType;
 import com.clouway.oauth2.user.IdentityFinder;
 import com.clouway.oauth2.user.User;
 import com.google.common.base.Optional;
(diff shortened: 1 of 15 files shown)


oauth2-example-app/src/main/java/com/clouway/oauth2/exampleapp/storage/InMemoryUserRepository.java
```diff
@@ -27,7 +28,7 @@
   }
 
   @Override
-  public Optional<Identity> findIdentity(String identityId, DateTime instantTime) {
+  public Optional<Identity> findIdentity(String identityId, GrantType grantType, DateTime instantTime) {
```
tsony-tsonev, line 31: Here at the implementation in the example app we could make a simple usage of the grantType with the Identity just to show its purpose and make the idea clearer.

tsony-tsonev (approved):
LGTM

oauth2-example-app/src/main/java/com/clouway/oauth2/exampleapp/storage/InMemoryUserRepository.java
mgenov: Good catch.

mgenov: I'll add.

mgenov: Added.

PR #24, opened 2016-10-13
oauth2: support multiple redirect urls per client

### LINKED ISSUE #23: oauth2: multiple redirect urls per client
Support for multiple redirect urls need to be added to provide a way to specify multiple endpoints for a single client.

This configuration will be mostly used for development environment where such configuration is necessary.

cc @mlesikov  @georgievJon

mgenov:
Added support for multiple redirect urls per client. When client sends authorization request to the oauth2 provider it may request and redirect_uri which to be preferred for the callback redirect. If it's not specified then the first one of the list is used.

In cases when the requested URL is fake the oauth2 provider will return bad request with proper information message to indicate that redirectUrl was broken.

Fixes #23

### DIFF
--- a/oauth2-example-app/src/main/java/com/clouway/oauth2/exampleapp/storage/InMemoryClientAuthorizationRepository.java
+++ b/oauth2-example-app/src/main/java/com/clouway/oauth2/exampleapp/storage/InMemoryClientAuthorizationRepository.java
@@ -45,9 +45,8 @@ class InMemoryClientAuthorizationRepository implements ClientAuthorizationReposi
   @Override
   public Optional<Authorization> authorize(Client client, String identityId, String responseType) {
     String code = tokenGenerator.generate();
-    String redirectURI = client.redirectURI;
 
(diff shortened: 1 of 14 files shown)


mlesikov:
so, your idea is that every cleint to have collection of  allowed redirect urls and the request may contains "redirect_url" param, which have to be from the allowed, otherwise if the param exists in the request the request does not pass. Also if the param does not exists in the request, then the first allowed from the database is used. 
Maybe there should be migration of the existing entities?

mgenov:
This change is performed in the library which is not related to any entities. The apps which are using it may require such as latest changes are released to maven central.

mgenov:
About the change, yes. A single Client in the context of oauth2 will be able to have multiple redirect_uri addresses.

mlesikov:
yes, I mean in the apps that use it, of course.

Also we need an service for these clients registration.

LGTM

mlesikov (approved):

mgenov:
The library already provides interface about this, so it shouldn't be hard to be implemented.

PR #25, opened 2016-10-14
oauth2: split IdentityFinder into 2 interfaces

mgenov:
Currently the IdentityFinder was responsible for retrieving of IdentityId from the Request and for retrieving of Identity information from the persistence layer (memory, db and etc). As this tasks are not related a new interface ResourceOwnerIdentityFinder was introduced to separate them.

The new responsibilities now are:
- ResourceOwnerIdentityFinder retrieve identity id from request
- IdentityFinder -> retrieve identity information for the RO


stefan-dimitrov:
LGTM.

PR #26, opened 2016-10-14
oauth2: remove not used properties from Client object

mgenov:
Unused properties in the OAuth2 Client object are removed such as:
- name - duplicated with description
- url - not used by the application


stefan-dimitrov:
LGTM.

PR #28, opened 2017-03-04
oauth2-server: remove primitive obsession of access token

mgenov:
Access token is no longer primitive which allows removing of the TTL property as it's calculated by the BearerToken.

### DIFF
--- a/oauth2-example-app/src/main/java/com/clouway/oauth2/exampleapp/storage/InMemoryTokens.java
+++ b/oauth2-example-app/src/main/java/com/clouway/oauth2/exampleapp/storage/InMemoryTokens.java
@@ -3,8 +3,8 @@ package com.clouway.oauth2.exampleapp.storage;
 import com.clouway.oauth2.DateTime;
 import com.clouway.oauth2.Duration;
 import com.clouway.oauth2.client.Client;
-import com.clouway.oauth2.token.GrantType;
 import com.clouway.oauth2.token.BearerToken;
+import com.clouway.oauth2.token.GrantType;
 import com.clouway.oauth2.token.TokenGenerator;
 import com.clouway.oauth2.token.TokenResponse;
 import com.clouway.oauth2.token.Tokens;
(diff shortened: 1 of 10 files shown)


oauth2-example-app/src/main/java/com/clouway/oauth2/exampleapp/storage/InMemoryTokens.java
```diff
@@ -23,12 +23,10 @@
   private final Map<String, BearerToken> tokens = Maps.newHashMap();
   private final Map<String, String> refreshTokenToAccessToken = Maps.newHashMap();
   private final TokenGenerator tokenGenerator;
-  private Duration timeToLive;
 
   @Inject
   public InMemoryTokens(TokenGenerator tokenGenerator, Duration timeToLive) {
```
qnislavnachev: Duration parameter is not used

qnislavnachev (changes requested):

oauth2-example-app/src/main/java/com/clouway/oauth2/exampleapp/storage/InMemoryTokens.java
mgenov: Fixed.

oauth2-server/src/main/java/com/clouway/oauth2/DateTime.java
```diff
@@ -1,11 +1,12 @@
 package com.clouway.oauth2;
 
+import java.io.Serializable;
 import java.util.Date;
 
 /**
  * @author Miroslav Genov (miroslav.genov@clouway.com)
  */
-public final class DateTime {
+public final class DateTime implements Serializable {
```
qnislavnachev, line 9: Why this class implements Serializable ?

mgenov: As it acts as a primitive (wrapper of date) someone would be able to use it for serialization (memcached, datastore and etc).

qnislavnachev:
LGTM

PR #29, opened 2017-03-29
oauth2-server: additional information about token

qnislavnachev:
additional info added to token such as email and client id

Related to #7

### DIFF
--- a/oauth2-server/src/main/java/com/clouway/oauth2/TokenInfoController.java
+++ b/oauth2-server/src/main/java/com/clouway/oauth2/TokenInfoController.java
@@ -31,9 +31,12 @@ class TokenInfoController implements InstantaneousRequest {
     BearerToken token = possibleToken.get();
 
     JsonObject o = new JsonObject();
+    o.addProperty("azp", token.clientId);
+    o.addProperty("aud", token.clientId);
     o.addProperty("sub", token.identityId);
     o.addProperty("exp", token.expirationTimestamp());
     o.addProperty("expires_in", token.ttlSeconds(instantTime));
+    o.addProperty("email", token.email);
 
     if (!token.scopes.isEmpty()) {
       o.addProperty("scope", Joiner.on(" ").join(token.scopes));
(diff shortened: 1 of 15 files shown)


oauth2-server/src/main/java/com/clouway/oauth2/TokenInfoController.java
```diff
@@ -31,9 +31,12 @@ public Response handleAsOf(Request request, DateTime instantTime) {
     BearerToken token = possibleToken.get();
 
     JsonObject o = new JsonObject();
+    o.addProperty("azp", token.clientId);
```
mgenov, line 34: It would be better IdentityFinder to be taken in the TokenInfoController instead of Tokens implementation is it won't force change on clients.

mgenov (changes requested):

mgenov:
It would be better if issueToken method accepts Identity object instead of just identityId. In this case client implementations of tokens would be able to store in the persistence/cache layer all properties which latter to be returned in the token response. 

The IssueNewTokenActivity is taking identity
```java
Authorization authorization = possibleResponse.get();

TokenResponse response = tokens.issueToken(GrantType.AUTHORIZATION_CODE, client, authorization.identityId, authorization.scopes, instant);
if (!response.isSuccessful()) {
  return OAuthError.invalidRequest("Token cannot be issued.");
}
```

So the Identity could be taken here as authorization is performed only once whether information about token can be executed many times and oauth2-server will gain performance benefit.

qnislavnachev:
issueToken accepts Identity object instead string of identity id and no longer client should inject IdentityFinder in Tokens Implementation

oauth2-server/src/main/java/com/clouway/oauth2/jwt/JwtController.java
```diff
@@ -2,10 +2,7 @@
 
 import com.clouway.friendlyserve.Request;
 import com.clouway.friendlyserve.Response;
-import com.clouway.oauth2.BearerTokenResponse;
-import com.clouway.oauth2.DateTime;
-import com.clouway.oauth2.InstantaneousRequest;
-import com.clouway.oauth2.OAuthError;
+import com.clouway.oauth2.*;
```
mgenov: You have to fix your IDEA settings in the Code Style section: 

screen shot 2017-03-30 at 10 10 25 am

Make "Class count to use *" 12 
Make "Name count to use static imports with *" 12

qnislavnachev:
IDEA settings fixed 
style

mgenov (approved):
LGTM !

Issue #12, opened 2016-06-08
openid: support id token for better performance


mgenov:
Additional information about the user that is authenticated/authorized could be attached when access token is generated. By encoding this information in the response client application will not be forced to send additional request to the server for retrieving of the user information. 

| Property | Description |
| --- | --- |
| **access_token** | A token that can be sent to the API server. |
| **id_token** | A JWT that contains identity information about the user that is digitally signed by Provider. |
| **token_type** | Identifies the type of token returned. At this time, this field always has the value Bearer. |
| **expires_in** | The remaining lifetime of the access token. |
| **refresh_token** | This field is only present if access_type=offline is included in the authentication request. |

So the following information could be encoded in the id_token attribute:

``` json
{"iss":"provider.somedomain.com",
 "at_hash":"HK6E_P6Dh8Y93mRNtsDB1Q",
 "email_verified":"true",
 "sub":"10769150350006150715113082367",
 "azp":"1234987819200.apps.somedomain.com",
 "email":"jsmith@example.com",
 "aud":"1234987819200.apps.somedomain.com",
 "iat":1353601026,
 "exp":1353604926,
 "hd":"example.com" }
```

Here is a sample response:

```
HTTP/1.1 200 OK
  Content-Type: application/json
  Cache-Control: no-store
  Pragma: no-cache

  {
   "access_token": "SlAV32hkKG",
   "token_type": "Bearer",
   "refresh_token": "8xLOxBtZp8",
   "expires_in": 3600,
   "id_token": "eyJhbGciOiJSUzI1NiIsImtpZCI6IjFlOWdkazcifQ.ewogImlzc
     yI6ICJodHRwOi8vc2VydmVyLmV4YW1wbGUuY29tIiwKICJzdWIiOiAiMjQ4Mjg5
     NzYxMDAxIiwKICJhdWQiOiAiczZCaGRSa3F0MyIsCiAibm9uY2UiOiAibi0wUzZ
     fV3pBMk1qIiwKICJleHAiOiAxMzExMjgxOTcwLAogImlhdCI6IDEzMTEyODA5Nz
     AKfQ.ggW8hZ1EuVLuxNuuIJKX_V8a_OMXzR0EHR9R6jgdqrOOF4daGU96Sr_P6q
     Jp6IcmD3HP99Obi1PRs-cwh3LO-p146waJ8IhehcwL7F09JdijmBqkvPeB2T9CJ
[...]

PR #30, opened 2017-03-29
oauth2: added id_token to bearer token response

VasilMitovClouway:
Added an id_token field to the response field. For the moment it is in String format, but I was wondering would it be a good idea to make an IdToken object with a method to be able to encode/decode it in base64 format. As I understand Tokens are made and added to the response from the repositories with the interface Tokens.
PS: I do not have rights to assign reviewers nor put labels
related #12

### DIFF
--- a/oauth2-example-app/src/main/java/com/clouway/oauth2/exampleapp/storage/InMemoryClientRepository.java
+++ b/oauth2-example-app/src/main/java/com/clouway/oauth2/exampleapp/storage/InMemoryClientRepository.java
@@ -55,6 +55,12 @@ class InMemoryClientRepository implements ClientRegistry, ClientFinder, ClientKe
     return null;
   }
 
+  @Override
+  public Map<String, Block> publicCertificates() {
+    return null;
+  }
+
+
   public void registerServiceAccount(String clientEmail, String privateKeyAsPem) {
     try {
       serviceAccountKeys.put(clientEmail, pem.parse(new ByteArrayInputStream(privateKeyAsPem.getBytes())));
--- a/oauth2-server/src/main/java/com/clouway/oauth2/IssueNewTokenActivity.java
+++ b/oauth2-server/src/main/java/com/clouway/oauth2/IssueNewTokenActivity.java
@@ -5,6 +5,10 @@ import com.clouway.friendlyserve.Response;
 import com.clouway.oauth2.authorization.Authorization;
 import com.clouway.oauth2.authorization.ClientAuthorizationRepository;
 import com.clouway.oauth2.client.Client;
+import com.clouway.oauth2.client.ClientKeyStore;
+import com.clouway.oauth2.jws.Pem;
+import com.clouway.oauth2.jws.Pem.Block;
+import com.clouway.oauth2.jwt.Jwt.Header;
 import com.clouway.oauth2.token.BearerToken;
 import com.clouway.oauth2.token.GrantType;
 import com.clouway.oauth2.token.TokenResponse;
@@ -12,21 +16,30 @@ import com.clouway.oauth2.token.Tokens;
 import com.clouway.oauth2.user.IdentityFinder;
 import com.google.common.base.Optional;
 
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.Random;
+
+import static com.clouway.oauth2.idtoken.IdTokenBuilder.newIdToken;
+import static com.clouway.oauth2.jwt.Jwt.ClaimSet.newClaimSet;
+
 /**
  * IssueNewTokenActivity is representing the activity which is performed for issuing of new token.
  *
  * @author Miroslav Genov (miroslav.genov@clouway.com)
  */
 class IssueNewTokenActivity implements ClientActivity {
-
   private final Tokens tokens;
   private final IdentityFinder identityFinder;
   private final ClientAuthorizationRepository clientAuthorizationRepository;
+  private final ClientKeyStore keyStore;
 
-  IssueNewTokenActivity(Tokens tokens, IdentityFinder identityFinder, ClientAuthorizationRepository clientAuthorizationRepository) {
+  IssueNewTokenActivity(Tokens tokens, IdentityFinder identityFinder, ClientAuthorizationRepository clientAuthorizationRepository, ClientKeyStore keyStore) {
     this.tokens = tokens;
     this.identityFinder = identityFinder;
     this.clientAuthorizationRepository = clientAuthorizationRepository;
+    this.keyStore = keyStore;
   }
 
   @Override
@@ -47,6 +60,10 @@ class IssueNewTokenActivity implements ClientActivity {
     }
 
     Identity identity = possibleIdentity.get();
+    Map<String, Block> certificates = keyStore.publicCertificates();
+    String certKey = randomKey(certificates);
+    Pem.Block key = certificates.get(certKey);
+
 
     TokenResponse response = tokens.issueToken(GrantType.AUTHORIZATION_CODE, client, identity, authorization.scopes, instant);
     if (!response.isSuccessful()) {
@@ -54,7 +71,27 @@ class IssueNewTokenActivity implements ClientActivity {
     }
 
     BearerToken accessToken = response.accessToken;
+    String host = request.header("Host");
 
-    return new BearerTokenResponse(accessToken.value, accessToken.ttlSeconds(instant), response.refreshToken);
+    String idToken = newIdToken().header(new Header("RS256"))
+            .claims(newClaimSet()
+                    .iss(host)
+                    .aud(client.id)
+                    .sub(identity.id())
+                    .iat(instant.timestamp())
+                    .exp(accessToken.expirationTimestamp())
+                    .certId(certKey)
+                    .customClaims(identity.claims())
+                    .build())
+            .signWith(key)
+            .build();
+
+    return new BearerTokenResponse(accessToken.value, accessToken.ttlSeconds(instant), response.refreshToken, idToken);
+  }
+
+  private String randomKey(Map<String, Block> map) {
+    Random random = new Random();
+    List<String> certId = new ArrayList<>(map.keySet());
+    return certId.get(random.nextInt(certId.size()));
   }
 }
--- a/oauth2-server/src/main/java/com/clouway/oauth2/idtoken/IdTokenBuilder.java
+++ b/oauth2-server/src/main/java/com/clouway/oauth2/idtoken/IdTokenBuilder.java
@@ -1,0 +1,47 @@
+package com.clouway.oauth2.idtoken;
+
+
+import com.clouway.oauth2.jws.Pem;
+import com.clouway.oauth2.jws.RsaJwsSignature;
+import com.clouway.oauth2.jwt.Jwt.ClaimSet;
+import com.clouway.oauth2.jwt.Jwt.Header;
+import com.google.common.io.BaseEncoding;
+
+/**
+ * @author Vasil Mitov <v.mitov.clouway@gmail.com>
+ */
+public class IdTokenBuilder {
+
+  private Header header;
+  private ClaimSet claims;
+  private Pem.Block key;
+
+  public static IdTokenBuilder newIdToken() {
+    return new IdTokenBuilder();
+  }
+
+  public IdTokenBuilder header(Header header) {
+    this.header = header;
+    return this;
+  }
+
+  public IdTokenBuilder claims(ClaimSet claims) {
+    this.claims = claims;
+    return this;
+  }
+
+  public IdTokenBuilder signWith(Pem.Block key) {
+    this.key = key;
+    return this;
+  }
+
+  public String build() {
+    String headerString = BaseEncoding.base64Url().omitPadding().encode(header.toString().getBytes());
+    String claimsString = BaseEncoding.base64Url().omitPadding().encode(claims.toString().getBytes());
+
+    String unsignedToken = String.format("%s.%s", headerString, claimsString);
+    RsaJwsSignature signature = new RsaJwsSignature(unsignedToken.getBytes());
+
+    return String.format("%s.%s", unsignedToken, signature.sign(unsignedToken.getBytes(), key));
+  }
+}
--- a/oauth2-server/src/main/java/com/clouway/oauth2/jws/Signature.java
+++ b/oauth2-server/src/main/java/com/clouway/oauth2/jws/Signature.java
@@ -1,5 +1,10 @@
 package com.clouway.oauth2.jws;
 
+import com.clouway.oauth2.jws.Pem.Block;
+
+import java.security.Key;
+import java.security.KeyPair;
+
 /**
  * Signature is representing a single JWS signature that is applied over received messages.
  * <p/>
@@ -12,10 +17,30 @@ public interface Signature {
    * Verify is verifying Signature using the provided privateKey as PEM file.
    * <p/>
    *
-   * @param content the content to be verified
+   * @param content    the content to be verified
    * @param privateKey the private key used for verifying
    * @return true if signature is
    */
-  boolean verify(byte[] content, Pem.Block privateKey);
+  boolean verifyWithPrivateKey(byte[] content, Pem.Block privateKey);
+
+  /**
+   * Verify is veryfying Signature using the provided publicKey as PEM file
+   * <p/>
+   *
+   * @param content to be verified
+   * @param key The public key to verify the content with
+   * @return
+   */
+  boolean verify(byte[] content, Block key);
+
+  /**
+   * Sign a token with a private key
+   * <p/>
+   *
+   * @param content of the token
+   * @param privateKey to sign the token with
+   * @return token signature to be added to the token header.claims
+   */
+  String sign(byte[] content, Pem.Block privateKey);
 
 }
--- a/oauth2-server/src/main/java/com/clouway/oauth2/jwt/Jwt.java
+++ b/oauth2-server/src/main/java/com/clouway/oauth2/jwt/Jwt.java
@@ -2,6 +2,8 @@ package com.clouway.oauth2.jwt;
 
 import com.google.common.base.MoreObjects;
 
+import java.util.Map;
+
 /**
  * Jwt stands for JSON Web Token and is the entry point of the JWT package.
  *
@@ -21,7 +23,7 @@ public final class Jwt {
     public final String alg;
 
     @SuppressWarnings("unused")
-    Header() {
+    public Header() {
       this(null);
     }
 
@@ -77,12 +79,90 @@ public final class Jwt {
      */
     public final String prn;
 
-    @SuppressWarnings("unused")
-    ClaimSet() {
-      this(null, null, null, null, null, null, null);
+    public final String certId;
+
+    public final Map<String, Object> customClaims;
+
+    public static final class Builder {
+      private String iss;
+      private String scope;
+      private String aud;
+      private Long exp;
+      private Long iat;
+      private String typ;
+      private String sub;
+      private String prn;
+      private String certId;
+      private Map<String, Object> customClaims;
+
+      private Builder() {
+      }
+
+      public ClaimSet build() {
+        return new ClaimSet(this);
+      }
+
+      public Builder iss(String iss) {
+        this.iss = iss;
+        return this;
+      }
+
+      public Builder scope(String scope) {
+        this.scope = scope;
+        return this;
+      }
+
+      public Builder aud(String aud) {
+        this.aud = aud;
+        return this;
+      }
+
+      public Builder exp(Long exp) {
+        this.exp = exp;
+        return this;
+      }
+
+      public Builder iat(Long iat) {
+        this.iat = iat;
+        return this;
+      }
+
+      public Builder typ(String typ) {
+        this.typ = typ;
+        return this;
+      }
+
+      public Builder sub(String sub) {
+        this.sub = sub;
+        return this;
+      }
+
+      public Builder prn(String prn) {
+        this.prn = prn;
+        return this;
+      }
+
+      public Builder certId(String certId) {
+        this.certId = certId;
+        return this;
+      }
+
+      public Builder customClaims(Map<String, Object> customClaims) {
+        this.customClaims = customClaims;
+        return this;
+      }
     }
 
-    public ClaimSet(String iss, String scope, String aud, Long exp, Long iat, String typ, String sub) {
+    public static Builder newClaimSet() {
+      return new Builder();
+    }
+
+    @SuppressWarnings("unused")
+    ClaimSet() {
+      this(null, null, null, null, null, null, null, null, null);
+    }
+
+    public ClaimSet(String iss, String scope, String aud, Long exp, Long iat, String typ, String sub, String certId, Map<String, Object> customClaims) {
       this.iss = iss;
       this.scope = scope;
       this.aud = aud;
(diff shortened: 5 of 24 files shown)


oauth2-example-app/src/main/java/com/clouway/oauth2/app/HttpClient.java
```diff
@@ -0,0 +1,65 @@
...
+import okhttp3.OkHttpClient;
+import okhttp3.Request;
+import okhttp3.Response;
+
+import java.io.IOException;
+
+/**
+ * @author Vasil Mitov <v.mitov.clouway@gmail.com>
+ */
+public class HttpClient {
```
mgenov: JwtClient

oauth2-example-app/src/main/java/com/clouway/oauth2/exampleapp/OauthAuthorizationServerModule.java
```diff
@@ -122,7 +122,7 @@ protected void doGet(HttpServletRequest req, HttpServletResponse resp) throws Se
             String token = req.getHeader("Authorization").split("\\s")[0];
```
mgenov: String token is no longer required as servlet prints the whole header.

oauth2-server/src/main/java/com/clouway/oauth2/token/TokenResponse.java
```diff
@@ -13,11 +13,13 @@
 
   public final BearerToken accessToken;
   public final String refreshToken;
+  public final String encodedIdToken;
```
mgenov: idToken is always encoded, so encoded doesn't give any additional value.

oauth2-server/src/test/java/com/clouway/oauth2/IssueNewTokenForClientTest.java
```diff
@@ -52,7 +51,7 @@ public void happyPath() throws IOException {
       will(returnValue(Optional.of(new Authorization("", "", "::user_id::", "::auth_code::", Collections.<String>emptySet(), Collections.singleton("::redirect_uri::")))));
 
       oneOf(tokens).issueToken(GrantType.AUTHORIZATION_CODE, client, "::user_id::", Collections.<String>emptySet(), anyTime);
-      will(returnValue(new TokenResponse(true, aNewToken().withValue("::token::").build(), "::refresh token::")));
+      will(returnValue(new TokenResponse(true, aNewToken().withValue("::token::").build(), "::refresh token::", "encodedIdToken")));
```
mgenov: ::id token::" to follow the existing pattern in code.

oauth2-server/src/test/java/com/clouway/oauth2/SerializeBearerTokensTest.java
```diff
@@ -17,26 +17,39 @@
...
+            "0KICJoZCI6ImV4YW1wbGUuY29tIiB9";
+
     String expectedResponse = "Content-Type: application/json; charset=utf-8\r\n" +
             "{\"access_token\":\"mF_9.B5f-4.1JqM\"," +
             "\"token_type\":\"Bearer\"," +
             "\"expires_in\":3600," +
-            "\"refresh_token\":" +
-            "\"tGzv3JOkF0XG5Qx2TlKWIA\"}";
+            "\"refresh_token\":\"tGzv3JOkF0XG5Qx2TlKWIA\"," +
+            "\"id_token\":\""+encodedIdToken+"\"}";
```
mgenov: idToken

oauth2-server/src/test/java/com/clouway/oauth2/SerializeBearerTokensTest.java
```diff
@@ -17,26 +17,39 @@
 
   @Test
   public void happyPath() throws IOException {
+    String encodedIdToken="eyJpc3MiOiJwcm92aWRlci5zb21lZG9tYWluLmNvbSIsDQogImF0X2hhc2giOiJISzZFX1A2RGg4WTkzbVJOdHNEQj" +
```
mgenov: You can use ::id token:: as value as it's value is not relevant with the test. You are just verifying that there is a property id_token and it's value is added to the response.

mgenov:
> Added an id_token field to the response field. For the moment it is in String format, but I was wondering would it be a good idea to make an IdToken object with a method to be able to encode/decode it in base64 format. As I understand Tokens are made and added to the response from the repositories with the interface Tokens.

If you add jjwt to be part of the project it won't be required: https://github.com/jwtk/jjwt

This is an example for building of JWT token using jjwt. 
```java
String compactJws = Jwts.builder()
  .setSubject("Joe")
  .signWith(SignatureAlgorithm.HS512, key)
  .compact();
```

Looks really simple and good but this library depends on jackson where we already use GSON in a lot of places. 
https://github.com/jwtk/jjwt/blob/master/pom.xml#L80

But you can use api as example point of building of JWT tokens.

mgenov (changes requested):

oauth2-example-app/src/main/java/com/clouway/oauth2/app/HttpClient.java
VasilMitovClouway: Renamed to JwtClient.

oauth2-example-app/src/main/java/com/clouway/oauth2/exampleapp/OauthAuthorizationServerModule.java
VasilMitovClouway: Removed unnecessary token.

oauth2-server/src/main/java/com/clouway/oauth2/token/TokenResponse.java
VasilMitovClouway: renamed to idToken.

oauth2-server/src/test/java/com/clouway/oauth2/IssueNewTokenForClientTest.java
VasilMitovClouway: "encodedToken" renamed to "::id token::" to follow existing testing pattern code.

oauth2-server/src/test/java/com/clouway/oauth2/SerializeBearerTokensTest.java
VasilMitovClouway: Removed unnecessary token string and replaced with ::id token::

oauth2-server/src/test/java/com/clouway/oauth2/SerializeBearerTokensTest.java
VasilMitovClouway: renamed to "::id token::"

VasilMitovClouway:
I could just add the jjwt library to the exampleapp pom and make an ExampleIdToken object, or ExampleIdTokenBuilder that returns a string, with documentation so that future users of our library can see an example way to create their own idTokens.

mgenov:
The example app doesn't have to create such. The oauth2-server should create them whether the example app should parse them.

VasilMitovClouway:
What algorithm should I use to sign the ID token, or should it be unsigned and just base64 encoded.

mgenov:
RS256

```java
    final SignatureFactory signatureFactory = new SignatureFactory() {
      @Override
      public Optional<Signature> createSignature(byte[] signatureValue, Header header) {
        // It's vulnerable multiple algorithms to be supported, so server need to reject
        // any calls from not supported algorithms
        // More infromation could be taken from: https://auth0.com/blog/2015/03/31/critical-vulnerabilities-in-json-web-token-libraries/
        if (!"RS256".equals(header.alg)) {
          return Optional.absent();
        }
        return Optional.<Signature>of(new RsaJwsSignature(signatureValue));
      }
    };
```

mgenov:
Few notes about the public keys.

The identity provider (accounts.yourappatyourdomain.com) should be able to serve list of several public keys which are used and for signing of ID tokens which means that they are global and should not be related with the concrete service account or client. 

Each app which uses the library should be able to provide it's own list of keys for signing and some of them to be used randomly from list. 

Added related issue about this functionality: #38

VasilMitovClouway:
May I ask for the Public key that corresponds to the Private key that we have in the project. I tried generating it trough the console but the JavaSecurity didn't like the output.
P.S I've tried using online generated RSA key pairs, alas JavaSecurity/Pem parser didn't like them as well.

mgenov:
You can generate your own key pairs for testing purposes.

mgenov:
Here is an example how to generate keys, encode/decode in pem, sign data and verify that data was signed by the owner: 

```java
 @Test
  public void anExampleHowToSignDataAndVerifyIt() throws Exception {
    KeyPairGenerator keyGen = KeyPairGenerator.getInstance("RSA");
    keyGen.initialize(2048);
    KeyPair keyPair = keyGen.generateKeyPair();

    PublicKey publicKey = keyPair.getPublic();
    PrivateKey privateKey = keyPair.getPrivate();

    Pem pem = new Pem();
    String publicKeyPem = pem.format(new Block("PUBLIC KEY", Collections.<String, String>emptyMap(), publicKey.getEncoded()));
    String privateKeyPem = pem.format(new Block("PRIVATE KEY", Collections.<String, String>emptyMap(), privateKey.getEncoded()));

    System.out.println(publicKeyPem);
    System.out.println(privateKeyPem);

    Pem.Block publicKeyBlock = pem.parse(new ByteArrayInputStream(publicKeyPem.getBytes()));
    byte[] publicKeyBytes = publicKeyBlock.getBytes();

    System.out.println("Public Key Matches: " + Arrays.equals(publicKey.getEncoded(), publicKeyBytes));

    byte[] content = "Hello, how are you ?".getBytes();
    
    byte[] signature = signBytes(privateKey, content);
    String encodedSignature = BaseEncoding.base64().encode(signature);

    byte[] decodedSignature = BaseEncoding.base64().decode(encodedSignature);
    System.out.println("Is this data signed by owner of this key: " + isSignatureValid(publicKey, content, decodedSignature));
  }

  private boolean isSignatureValid(PublicKey key,byte[] data, byte[] signature) throws NoSuchAlgorithmException, InvalidKeyException, SignatureException {
    java.security.Signature sig = java.security.Signature.getInstance("SHA256withRSA");
    sig.initVerify(key);
    sig.update(data);

    return sig.verify(signature);
  }

  private byte[] signBytes(PrivateKey privateKey, byte[] content) throws InvalidKeyException, NoSuchAlgorithmException, SignatureException {
    java.security.Signature sig = java.security.Signature.getInstance("SHA256withRSA");
    sig.initSign(privateKey);
    sig.update(content);
    return sig.sign();
  }
```

VasilMitovClouway:
Update: If I sign it with the private key and verify it with it, the test passes.

mgenov:
As you can see in my example, signature is created with private key whether verification is performed with public key.

mgenov:
The following snippet works as expected: https://gist.github.com/mgenov/ce13d19e76c2b10a500adb82ea456942


Few other notes: 
 * ClaimSet is already defined in the JWT package: Jwt.Header, Jwt.ClaimSet. 
 * The id_token should has the same information as returned one from the UserInfoController: id, name, email, given_name, family_name and custom claims from Identity

VasilMitovClouway:
Meaning that i shoud rename the current IdToken.ClaimSet, and give it fields: id, name, email, given_name, family_name and custom_claims ?

mgenov:
If the existing Jwt.ClaimSet cannot be used and for the id token then it should stay or may wrap the existing claim set. I suppose they should be similar so one of them should do the job.

http://openid.net/specs/openid-connect-core-1_0.html#id_tokenExample
Here is another id_token example:
```json
{
 "iss": "http://server.example.com",
 "sub": "248289761001",
 "aud": "s6BhdRkqt3",
 "nonce": "n-0S6_WzA2Mj",
 "exp": 1311281970,
 "iat": 1311280970,
 "name": "Jane Doe",
 "given_name": "Jane",
 "family_name": "Doe",
 "gender": "female",
 "birthdate": "0000-10-31",
 "email": "janedoe@example.com",
 "picture": "http://example.com/janedoe/me.jpg"
}
```

mgenov:
please rebase your changes with upstream cause i'm seeing changes and from other PR.

oauth2-server/src/main/java/com/clouway/oauth2/idtoken/IdToken.java
```diff
@@ -0,0 +1,123 @@
+package com.clouway.oauth2.idtoken;
+
+import com.clouway.oauth2.jwt.Jwt;
+
+/**
+ * @author Vasil Mitov <v.mitov.clouway@gmail.com>
+ */
+public final class IdToken {
```
mgenov: IdToken has zero fields and I'm not sure that you need it.

mgenov (changes requested):

oauth2-server/src/main/java/com/clouway/oauth2/idtoken/IdToken.java
VasilMitovClouway: It's corresponding Claims has different fields than the JWT. Was wondering if I can just change the JWT ClaimSet fields and add the ones I need for the Token to have the needed fields.

mgenov: ClaimSet  may have a custom claims represented as Map<String, Object>, so everyone will be able to add them if it's required.

oauth2-server/src/test/java/com/clouway/oauth2/PemKeyGenerator.java
```diff
@@ -0,0 +1,55 @@
...
+    } catch (IOException e) {
+      e.printStackTrace();
+    }
+    return null;
+  }
+
+  public static Pem.Block generatePrivateKey() {
+    Pem pem = new Pem();
+    PrivateKey privateKey = keyPair.getPrivate();
+    String publicKeyPem = pem.format(new Block("PRIVATE KEY", Collections.<String, String>emptyMap(), privateKey.getEncoded()));
```
mgenov: privateKeyPem 


Actually you don't need format/parse, as you can return Pem.Block object directly. 
```java
new Pem.Block("PRIVATE KEY", Collections.<String, String>emptyMap(), privateKey.getEncoded());
```

oauth2-server/src/main/java/com/clouway/oauth2/idtoken/IdToken.java
VasilMitovClouway: IdToken removed, now the JWT is used in the IdTokenBuilder.

oauth2-example-app/src/main/java/com/clouway/oauth2/exampleapp/storage/InMemoryClientRepository.java
```diff
@@ -55,6 +55,16 @@ public Client register(ClientRegistrationRequest request) {
     return null;
   }
 
+  @Override
+  public Map<String, Block> certificates() {
```
mgenov: publicCertificates

oauth2-example-app/src/main/java/com/clouway/oauth2/exampleapp/storage/InMemoryClientRepository.java
```diff
@@ -55,6 +55,16 @@ public Client register(ClientRegistrationRequest request) {
...
+    return null;
+  }
+
+  @Override
+  public Optional<Block> getRandomPrivateKey() {
```
mgenov: idTokenSigningKey()

oauth2-server/src/main/java/com/clouway/oauth2/IssueNewTokenActivity.java
```diff
@@ -55,6 +73,40 @@ public Response execute(Client client, Request request, DateTime instant) {
...
+                    .certId(certId)
+                    .customClaims(identity.claims())
+                    .build())
+            .signWith(key)
+            .build();
+
+    return new BearerTokenResponse(accessToken.value, accessToken.ttlSeconds(instant), response.refreshToken, idToken);
+  }
+
+  private Entry<String, Block> randomIdentifierKeyPair() {
```
mgenov: Looks complicated and could be simplified with something like:
```java
Random random = new Random();
List<String> keys  = new ArrayList<>(certificates.keySet());
String   randomKey = keys.get(random.nextInt(keys.size()));
return certificates.get(randomKey);
```

oauth2-server/src/test/java/com/clouway/oauth2/IdTokenTest.java
```diff
@@ -0,0 +1,77 @@
...
+                  .sub("10769150350006150715113082367")
+                  .iat(123456789L)
+                  .customClaims(customClaims)
+                  .build())
+          .signWith(privateKey)
+          .build();
+
+  @Test
+  public void tokenVerifiedWithPrivateKey() throws Exception {
+    byte[] signatureValue = BaseEncoding.base64().decode(token.split("\\.")[2]);
```
mgenov:
```java
String[] parts = token.split("\\.");
byte[] signature = BaseEncoding.base64().decode(parts[2]);
String value = String.format("%s.%s", parts[0], parts[1]); 
```

oauth2-server/src/test/java/com/clouway/oauth2/IdTokenTest.java
```diff
@@ -0,0 +1,77 @@
...
+import static com.clouway.oauth2.PemKeyGenerator.generatePublicKey;
+import static com.clouway.oauth2.idtoken.IdTokenBuilder.newIdToken;
+import static com.clouway.oauth2.jwt.Jwt.ClaimSet.newClaimSet;
+import static org.hamcrest.CoreMatchers.is;
+import static org.junit.Assert.*;
+
+/**
+ * @author Vasil Mitov <v.mitov.clouway@gmail.com>
+ */
+public class IdTokenTest {
```
mgenov: VerifyIdTokenTest

oauth2-server/src/main/java/com/clouway/oauth2/IssueNewTokenActivity.java
```diff
@@ -5,28 +5,42 @@
...
  *
  * @author Miroslav Genov (miroslav.genov@clouway.com)
  */
 class IssueNewTokenActivity implements ClientActivity {
-
   private final Tokens tokens;
   private final IdentityFinder identityFinder;
   private final ClientAuthorizationRepository clientAuthorizationRepository;
+  private final ClientKeyStore keyStore;
+  private final String clientUrl;
```
mgenov: I've though about this and the additional clientUrl option could be skipped by using the Host header from the request.

The following line:
```java
String host = request.header("Host"); //case for name need to be checked: host Host, HOST
```
should return the target host which was called by the client (which is the iss property actually)

oauth2-server/src/main/java/com/clouway/oauth2/idtoken/IdTokenBuilder.java
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  public IdTokenBuilder signWith(Pem.Block key) {
+    this.key = key;
+    return this;
+  }
+
+  public String build() {
+    try {
+      String headerString = BaseEncoding.base64Url().omitPadding().encode(header.toString().getBytes());
```
mgenov: Looks duplicated with RsaJwsSignature.

mgenov (changes requested):

oauth2-example-app/src/main/java/com/clouway/oauth2/exampleapp/storage/InMemoryClientRepository.java
VasilMitovClouway: changed to publicCertificates

oauth2-server/src/main/java/com/clouway/oauth2/IssueNewTokenActivity.java
VasilMitovClouway: Removed clientUrl, now iss is taken from the request.header() method.

oauth2-server/src/main/java/com/clouway/oauth2/IssueNewTokenActivity.java
VasilMitovClouway: Simplified method but still using entry because I need the Key ID as well so that when the token is verified it will be simpler/easier.

oauth2-server/src/test/java/com/clouway/oauth2/IdTokenTest.java
VasilMitovClouway: Test name changed

oauth2-server/src/main/java/com/clouway/oauth2/idtoken/IdTokenBuilder.java
VasilMitovClouway: Removed duplication.

oauth2-server/src/main/java/com/clouway/oauth2/jwt/Jwt.java
```diff
@@ -91,12 +97,102 @@ public ClaimSet(String iss, String scope, String aud, Long exp, Long iat, String
...
 
     @Override
     public String toString() {
-      return MoreObjects.toStringHelper(this).add("iss", iss).add("scope", scope).add("aud", aud).add("exp", exp).add("iat", iat).add("typ", typ).toString();
+      return MoreObjects.toStringHelper(this).add("iss", iss).add("scope", scope).add("aud", aud).add("exp", exp)
+              .add("iat", iat).add("typ", typ).add("certificate_id", certId).add("custom_claims", customClaims).toString();
+    }
+
+
+    public static final class Builder {
```
mgenov: Builder should go first by convention. 

https://github.com/clouway/teamspace/wiki/Java-Code-Convention

oauth2-server/src/main/java/com/clouway/oauth2/jws/Signature.java
```diff
@@ -12,10 +17,14 @@
...
    * <p/>
    *
-   * @param content the content to be verified
+   * @param content    the content to be verified
    * @param privateKey the private key used for verifying
    * @return true if signature is
    */
   boolean verify(byte[] content, Pem.Block privateKey);
 
+  boolean verifyWithPublicKey(byte[] content, Block key);
```
mgenov: As verification with private key (verify method) is specific it would be better: 

verify() to be renamed to: verifyWithPrivateKey 
verifyWithPublicKey to be renamed to just verify()

So the method signatures will look as follow: 

```java
// verify with public key
verify(byte[] content, Pem.Block publicKey);

// verify with private key
verifyWithPrivateKey(byte[] content, Pem.Block privateKey);

// sign with private key
sign(byte[] content, Pem.Block privateKey);

```

Comments are missing for the new methods.

mgenov (changes requested):
Overall things are looking fine. Just few minor changes need to be performed.

oauth2-server/src/main/java/com/clouway/oauth2/jwt/Jwt.java
VasilMitovClouway: Class rearranged to properly follow convention

oauth2-server/src/main/java/com/clouway/oauth2/jws/Signature.java
VasilMitovClouway: Renamed methods

oauth2-server/src/main/java/com/clouway/oauth2/jws/Signature.java
```diff
@@ -12,10 +17,14 @@
...
    *
-   * @param content the content to be verified
+   * @param content    the content to be verified
    * @param privateKey the private key used for verifying
    * @return true if signature is
    */
-  boolean verify(byte[] content, Pem.Block privateKey);
+  boolean verifyWithPrivateKey(byte[] content, Pem.Block privateKey);
+
+  boolean verify(byte[] content, Block key);
```
mgenov: Add comments to the new methods.

VasilMitovClouway: Just realised that i didnt add documentation. Documentation is now added.

mgenov: I'm not able to see it.

VasilMitovClouway: Which of the two documentations of the verify methods is better to leave?

mgenov: I don't understand you. Every interface should be well documented.

mgenov (approved):
LGTM !

Issue #32, opened 2017-03-30
oauth2/api: improve design of ClientRepository


mgenov:
The design of ClientRepository forces Client to be registered with already generated ID. Most database are providing mechanism for generation of ids and it would be better if the ClientRepository is responsible for the generation of the ids. 

The existing design is: 
```java 
public interface ClientRepository {

  void register(Client client) throws ClientAlreadyExistsException;

  Optional<Client> findById(String id);
}
```

but it would be better if it's changed to something like: 

```java
public interface ClientRepository {

  Client register(ClientRegistrationRequest client) throws ClientAlreadyExistsException;

  Optional<Client> findById(String id);
}
```

so the identifier of the client will be added by the persistence layer.

mgenov:
Added in #33

PR #33, opened 2017-03-30
oauth2-server: improved client repository

qnislavnachev:
no longer client repositry accepts Client Object, instead of that
it accept ClientRegistrationRequest that gives info about client
that will be stored
PS: I think  about that it can be added a seccond method that accepts
client object so that user can use both with Client or ClientRegistrationRequest

Related to #32

### DIFF
--- a/oauth2-example-app/src/test/java/com/clouway/oauth2/exampleapp/storage/InMemoryClientRepositoryTest.java
+++ b/oauth2-example-app/src/test/java/com/clouway/oauth2/exampleapp/storage/InMemoryClientRepositoryTest.java
@@ -1,17 +1,26 @@
 package com.clouway.oauth2.exampleapp.storage;
 
+import com.clouway.oauth2.client.Client;
+import com.clouway.oauth2.client.ClientAlreadyExistsException;
+import com.clouway.oauth2.client.ClientRegistrationRequest;
 import com.clouway.oauth2.client.ClientRepository;
 import com.clouway.oauth2.exampleapp.ClientRepositoryContractTest;
+import com.google.common.base.Optional;
 
 /**
  * @author Ivan Stefanov <ivan.stefanov@clouway.com>
  */
-public class InMemoryClientRepositoryTest extends ClientRepositoryContractTest {
+public class InMemoryClientRepositoryTest implements ClientRepository {
 
   private InMemoryClientRepository repository = new InMemoryClientRepository();
 
   @Override
-  public ClientRepository createRepository() {
-    return repository;
+  public Client register(ClientRegistrationRequest request) throws ClientAlreadyExistsException {
+    return repository.register(request);
+  }
+
+  @Override
+  public Optional<Client> findById(String id) {
+    return repository.findById(id);
   }
 }
--- a/oauth2-server/src/main/java/com/clouway/oauth2/client/ClientRepository.java
+++ b/oauth2-server/src/main/java/com/clouway/oauth2/client/ClientRepository.java
@@ -7,7 +7,7 @@ import com.google.common.base.Optional;
  */
 public interface ClientRepository {
 
-  void register(Client client) throws ClientAlreadyExistsException;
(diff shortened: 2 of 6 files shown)


oauth2-server/src/main/java/com/clouway/oauth2/client/Client.java
```diff
@@ -25,6 +25,13 @@ public Client(String id, String secret, String description, Set<String> redirect
     this.redirectURLs = redirectURLs;
   }
 
+  public Client(String id, ClientRegistrationRequest request) {
```
mgenov: Just pass them as properties. Client doesn't have to know about the request.

oauth2-example-app/src/test/java/com/clouway/oauth2/exampleapp/storage/InMemoryClientRepositoryTest.java
```diff
@@ -8,10 +8,8 @@
  */
 public class InMemoryClientRepositoryTest extends ClientRepositoryContractTest {
 
-  private InMemoryClientRepository repository = new InMemoryClientRepository();
-
   @Override
-  public ClientRepository createRepository() {
-    return repository;
+  public ClientRepository createRepository(String defaultClientId) {
```
mgenov: This signature doesn't look good. 

What is the goal of the defaultClientId ? Client is represented minimum with clientId and secret.

mgenov (changes requested):

oauth2-example-app/src/test/java/com/clouway/oauth2/exampleapp/storage/InMemoryClientRepositoryTest.java
qnislavnachev: the goal of that defaultClientId is that when i am going to test, if client is registered i need  to know his id to retrieve it from repository, so in other words I pass the id of client thats going to be registered in constructor to make the test easier

qnislavnachev: It looks a little strange to me too, I will think better way to do that

qnislavnachev:
ClientRepository.register is no longer void method, now it returns String(the id of user that was registered) for removing the bad signiture, and removed the client constructor with that create a client by ClientRegistrationRequest object

oauth2-server/src/main/java/com/clouway/oauth2/client/ClientRepository.java
```diff
@@ -7,7 +7,7 @@
  */
 public interface ClientRepository {
 
-  void register(Client client) throws ClientAlreadyExistsException;
+  String register(ClientRegistrationRequest client) throws ClientAlreadyExistsException;
```
mgenov: The proposed signature was Client to be returned.

mgenov: https://github.com/clouway/oauth2-server/issues/32

mgenov:
> ClientRepository.register is no longer void method, now it returns String(the id of user that was registered) for removing the bad signiture, and removed the client constructor with that create a client by ClientRegistrationRequest object

These are clients for apps, not for users.

qnislavnachev:
clientRepository.register changed to return client object

mgenov (approved):
LGTM ! 

After a bit of thinking, it doesn't sound good register method to be part of the library as it's related with the logic in the applications. 

In following PR this method could be removed and name of the interface to be changed to ClientFinder as it will be used by the library only for retrieving.

mgenov:
Please use Fixes #32  instead of Related as it will close issue automatically.

qnislavnachev:
So in that facts clientRepository.register should be removed and it should be implemented in the application is that correct?

mgenov:
Yeah, cause the application may add additional fields to the Client, where by using this design it won't be possible.

PR #36, opened 2017-03-30
oauth2-server: clientRepository renamed to clientFinder

### LINKED ISSUE #34: oauth2/api: clean up ClientRepository interface
The interface for clients is having method for registration of clients which is related for the identity provider himself and should be open by the library. This means that the existing ClientRepository interface could be renamed to ClientFinder and the existing register method to be completely removed from the library.

qnislavnachev:
Client repository renamed to ClientFinder register method is completely removed from the interface, now client repository is implemented in the application, so it can be customized, added some additional test in example app

Fixes #34


oauth2-example-app/src/main/java/com/clouway/oauth2/exampleapp/ClientRepository.java
```diff
@@ -0,0 +1,14 @@
+package com.clouway.oauth2.exampleapp;
+
+import com.clouway.oauth2.client.Client;
+import com.clouway.oauth2.client.ClientRegistrationRequest;
+
+/**
+ *  Created by IaNiTyy on 30.03.17.
+ */
+public interface ClientRepository {
```
mgenov: ClientRegistry would be better.

mgenov (changes requested):

qnislavnachev:
ClientRepository interface in example app changed to ClientRegistry

mgenov (approved):
LGTM ! 

File comments need to follow the project convention: "Created by IaNiTyy on 30.03.17.".

PR #42, opened 2017-04-10
oauth2-server: remove id_token property from TokenResponse

mgenov:
IdToken property is removed from the TokenResponse as it's generation should be responsibility of the library and not of the client apps which are using it.

Still refresh token handler and JWT handler are not using it as expected, but it will be added in following PR.

Related with #43


mgenov:
friendly reminder

VasilMitovClouway:
For now, we are leaving the idToken property in the BearerTokenResponce?

qnislavnachev:
I think we will broke the api if we made this changes, I think JWT to be fixed as expected.

VasilMitovClouway:
In SerializeBearerTokensTest there is still a property in the String response that could be misleading for other people.

mgenov:
We are leaving it from TokenResponse not from BearerTokenResponse. 

Added separate constructor in BearerTokenResponse which to be used by JWT and RefreshToken handlers till we add this info and there.

VasilMitovClouway:
LGTM! after removing property in SerializeBearerTokensTest

Issue #45, opened 2017-04-11
oauth2: id token should not be generated when there are no public certificates


mgenov:
The ID token should not be generated when there are no configured public certificates. 

The library is failing with Exception currently if there are no configured public certificates.

mgenov:
Fixed in #47

PR #47, opened 2017-04-12
oauth2-server: added certificate availability check

VasilMitovClouway:
Id_token no longer created in IssueNewTokenActivity if there are no certificates configured.
related #45

### DIFF
--- a/oauth2-server/src/main/java/com/clouway/oauth2/IssueNewTokenActivity.java
+++ b/oauth2-server/src/main/java/com/clouway/oauth2/IssueNewTokenActivity.java
@@ -43,29 +43,31 @@ class IssueNewTokenActivity implements AuthorizedClientActivity {
 
   @Override
   public Response execute(Client client, Identity identity, Set<String> scopes, Request request, DateTime instant) {
-    Map<String, Block> certificates = keyStore.privateCertificates();
-    String certKey = randomKey(certificates);
-    Pem.Block key = certificates.get(certKey);
-
     TokenResponse response = tokens.issueToken(GrantType.AUTHORIZATION_CODE, client, identity, scopes, instant);
     if (!response.isSuccessful()) {
       return OAuthError.invalidRequest("Token cannot be issued.");
     }
 
     BearerToken accessToken = response.accessToken;
-    String host = request.header("Host");
+    Map<String, Block> certificates = keyStore.privateCertificates();
+    String idToken = "";
 
-    String idToken = Jwts.builder()
-            .setHeaderParam("cid", certKey)//CertificateId - the ID of the certificate that the token was signed with.
-            .setIssuer(host)
-            .setAudience(client.id)
-            .setSubject(identity.id())
-            .setIssuedAt(instant.asDate())
-            .setExpiration(new Date(accessToken.expirationTimestamp()))
-            .claim("identity", identityAsJson(identity).toString())
-            .signWith(SignatureAlgorithm.RS256, parsePem(key))
-            .compact();
+    if (certificates != null && !certificates.isEmpty()) {
+      String certKey = randomKey(certificates);
+      Pem.Block key = certificates.get(certKey);
+      String host = request.header("Host");
 
+      idToken = Jwts.builder()
+              .setHeaderParam("cid", certKey)//CertificateId - the ID of the certificate that the token was signed with.
+              .setIssuer(host)
+              .setAudience(client.id)
+              .setSubject(identity.id())
+              .setIssuedAt(instant.asDate())
+              .setExpiration(new Date(accessToken.expirationTimestamp()))
+              .claim("identity", identityAsJson(identity).toString())
+              .signWith(SignatureAlgorithm.RS256, parsePem(key))
+              .compact();
+    }
     return new BearerTokenResponse(accessToken.value, accessToken.ttlSeconds(instant), response.refreshToken, idToken);
   }
 
--- a/oauth2-server/src/test/java/com/clouway/oauth2/IssueNewTokenForClientTest.java
+++ b/oauth2-server/src/test/java/com/clouway/oauth2/IssueNewTokenForClientTest.java
@@ -26,7 +26,8 @@ import static com.clouway.oauth2.authorization.AuthorizationBuilder.newAuthoriza
 import static com.clouway.oauth2.client.ClientBuilder.aNewClient;
 import static org.hamcrest.Matchers.containsString;
 import static org.hamcrest.Matchers.is;
-import static org.junit.Assert.assertThat;
+import static org.hamcrest.Matchers.not;
+import static org.junit.Assert.*;
 
 /**
  * @author Miroslav Genov (miroslav.genov@clouway.com)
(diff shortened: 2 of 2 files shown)


VasilMitovClouway:
Cannot add labels or assign reviewers.

oauth2-server/src/test/java/com/clouway/oauth2/IssueNewTokenForClientTest.java
```diff
@@ -94,20 +94,13 @@ public void tokenCannotBeIssued() throws Exception {
 
       oneOf(identityFinder).findIdentity(with(any(String.class)), with(any(GrantType.class)), with(any(DateTime.class)));
       will(returnValue(Optional.of(identity)));
-
```
mgenov: I'm not able to see test for the new behaviour.

mgenov (changes requested):

oauth2-server/src/test/java/com/clouway/oauth2/IssueNewTokenForClientTest.java
VasilMitovClouway:
```java
@Test
  public void publicCertificatesNotConfigured() throws Exception {
    IssueNewTokenActivity controller = new IssueNewTokenActivity(tokens, identityFinder, clientAuthorizationRepository, keyStore);
    final Client client = aNewClient().build();
    final DateTime anyTime = new DateTime();
    final Identity identity = aNewIdentity().withId("::user_id::").build();

    context.checking(new Expectations() {{
      oneOf(clientAuthorizationRepository).findAuthorization(with(any(Client.class)), with(any(String.class)));
      will(returnValue(Optional.of(new Authorization("", "", "::user_id::", "::auth_code::", Collections.<String>emptySet(), Collections.singleton("::redirect_uri::")))));

      oneOf(tokens).issueToken(GrantType.AUTHORIZATION_CODE, client, identity, Collections.<String>emptySet(), anyTime);
      will(returnValue(new TokenResponse(true, aNewToken().withValue("::token::").build(), "::refresh token::")));

      oneOf(identityFinder).findIdentity(with(any(String.class)), with(any(GrantType.class)), with(any(DateTime.class)));
      will(returnValue(Optional.of(identity)));

      oneOf(keyStore).publicCertificates();
      will(returnValue(Collections.emptyMap()));
    }});

    Response response = controller.execute(client, new ParamRequest(ImmutableMap.of("code", "::auth_code::", "redirect_uri", "::redirect_uri::")), anyTime);
    String body = new RsPrint(response).printBody();

    assertThat(body, containsString("::token::"));
  }
```

The thing is that this test is an absolute copy of the happy path with the difference that public certs return empty map and I didnt think it was informative enought to be added.

mgenov: It indicates a smell in the code as this check need to be performed here which actually need to be considered.

mgenov: Keep in mind that signing need to be performed by private keys and verification to be performed with public keys, which means that name of the method is misleading and requires some fixes.

mgenov: The second comment is not going to be addressed here unless you see it as reasonable to be here.

VasilMitovClouway:
Regarding the availability check, we had a discussion about the ClientKeyStore.certificates() to return a Certificate object. Here is an example implementation of the Certificate Object and a helper class CertificateParcer

```java
public class Certificate {
  public final String certId;
  public final String privatePem;
  public final String publicPem;

  public Certificate(String certId, String privatePem, String publicPem) {
    this.certId = certId;
    this.privatePem = privatePem;
    this.publicPem = publicPem;
  }

  @Override
  public boolean equals(Object o) {
    if (this == o) return true;
    if (o == null || getClass() != o.getClass()) return false;
    Certificate that = (Certificate) o;
    return Objects.equal(certId, that.certId) &&
            Objects.equal(privatePem, that.privatePem) &&
            Objects.equal(publicPem, that.publicPem);
  }

  @Override
  public int hashCode() {
    return Objects.hashCode(certId, privatePem, publicPem);
  }
}
```
```java
public class CertificateParcer {

  public static PrivateKey getPrivateKey(Certificate certificate) {
    Block block = getPemBlock(certificate.privatePem);
    return (PrivateKey) getKey("private",block);
  }

  public static PublicKey getPublicKey(Certificate certificate) {
    Block block = getPemBlock(certificate.privatePem);
    return (PublicKey) getKey("public",block);
  }

  public static Pem.Block getPrivatePem(Certificate certificate) {
    return getPemBlock(certificate.privatePem);
  }

  public static Pem.Block getPublicPem(Certificate certificate) {
    return getPemBlock(certificate.publicPem);
  }

  private static Key getKey(String type,Block block){
    try {
      PKCS8EncodedKeySpec keySpec = new PKCS8EncodedKeySpec(block.getBytes());
      KeyFactory kf = KeyFactory.getInstance("RSA");
      switch (type){
        case "private":return kf.generatePrivate(keySpec);
        case "public":return kf.generatePublic(keySpec);
      }
    } catch (NoSuchAlgorithmException|InvalidKeySpecException e) {
      e.printStackTrace();
    }
    return null;
  }

  private static Block getPemBlock(String pemString) {
    Pem pem = new Pem();
    try {
      Pem.Block block = pem.parse(new ByteArrayInputStream(pemString.getBytes()));
      return block;
    } catch (IOException e) {
      e.printStackTrace();
    }
    return null;
  }
}
```

mgenov:
Maybe it would be better client apps to pass PrivateKey and PublicKey objects directly instead of PEM. This will let them decide in what form to be persisted. 

About this design, it's not visible how this code is going to be used by the IssueNewTokenActivity. The problem there is that it has many dependencies which causes tests to blow.

mgenov:
About the Certificate Class: 

Keep in mind that Java already has such classes:
https://docs.oracle.com/javase/7/docs/api/java/security/cert/Certificate.html
https://docs.oracle.com/javase/7/docs/api/java/security/cert/X509Certificate.html

and there is a difference between private/public key and certificate (certificate adds additional information about the owner of the certificate). It would be better library to support certificates instead of simple keys.

Take a look at: 
https://www.googleapis.com/oauth2/v1/certs

As you can see it serves CERTIFICATES, not keys.

VasilMitovClouway:
I think now the tests are more readable. Moved the keystore.privateCertificates() a bit lower, because it is not needed to be done when Oauth.error is returned shortening a few tests and it is now easier to read the tests when certificates are not returned after #55 .

oauth2-server/src/test/java/com/clouway/oauth2/IssueNewTokenForClientTest.java
```diff
@@ -65,6 +65,27 @@ public void happyPath() throws IOException {
   }
 
   @Test
+  public void idTokenNotReturnedWhenCertificatesNotImplemented() throws Exception {
```
mgenov: noPrivateCertsAreConfigured

oauth2-server/src/main/java/com/clouway/oauth2/client/CertificatesNotConfiguredException.java
```diff
@@ -0,0 +1,8 @@
+package com.clouway.oauth2.client;
+
+/**
+ * @author Vasil Mitov <v.mitov.clouway@gmail.com>
+ */
+public class CertificatesNotConfiguredException extends RuntimeException {
```
mgenov: This exception is not used.

oauth2-server/src/main/java/com/clouway/oauth2/IssueNewTokenActivity.java
```diff
@@ -43,29 +43,31 @@
...
-            .setIssuedAt(instant.asDate())
-            .setExpiration(new Date(accessToken.expirationTimestamp()))
-            .claim("identity", identityAsJson(identity).toString())
-            .signWith(SignatureAlgorithm.RS256, parsePem(key))
-            .compact();
-
+    Map<String, Block> certificates = keyStore.privateCertificates();
+    String idToken = "";
+
+    if (available(certificates)) {
```
mgenov: I'm not sure that null check need to be added. It's bad practices methods to return null but we cannot relay on correct behaviour from client apps. 

it's more readable if it's inline:

```java
if (certificates != null && !certificates.isEmpty()) {

}
```

mgenov (changes requested):

oauth2-server/src/test/java/com/clouway/oauth2/IssueNewTokenForClientTest.java
VasilMitovClouway: Tried to be as specific as I can in the description. Name is now changed

oauth2-server/src/main/java/com/clouway/oauth2/client/CertificatesNotConfiguredException.java
VasilMitovClouway: Removed

oauth2-server/src/main/java/com/clouway/oauth2/IssueNewTokenActivity.java
VasilMitovClouway: Removed private method and moved the checks inline. We have tested this if block, and it fixes the bug that occurred last week

mgenov (approved):
LGTM !

Issue #48, opened 2017-04-12
oauth2: improve design of RsaJwsSignature 


mgenov:
The existing RsaJwsSignature was designed just for verification but as extension of the library to support signing (id tokens) it requires re-design. 

```java
public class RsaJwsSignature implements com.clouway.oauth2.jws.Signature {
  private byte[] signature;

  public RsaJwsSignature(byte[] signature) {
    this.signature = signature;
  }

  public boolean verifyWithPrivateKey(byte[] content, Pem.Block privateKey) {
     ...
  }
  @Override
  public boolean verify(byte[] content, Block key) {
      ..
  }

  @Override
  public String sign(byte[] content, Pem.Block privateKey) {
      ... 
  }
}
```

The Constructor accepts signature  but in signing process it's not required as it's the result. This is why the design of sign/verify need to be improved.

mgenov:
The design is going to be changed with adding of #50

VasilMitovClouway:
Fixed in #51

PR #51, opened 2017-04-13
oauth2-server: split verify and sign functionality

VasilMitovClouway:
Added a RsaJwsSigner with a static method sign for RSA signing with
Pem.Blocks and removed the signing functionality from RsaJwsSignature, which now only verifies the tokens.
related #48

### DIFF
--- a/oauth2-server/src/main/java/com/clouway/oauth2/idtoken/IdTokenBuilder.java
+++ b/oauth2-server/src/main/java/com/clouway/oauth2/idtoken/IdTokenBuilder.java
@@ -2,10 +2,12 @@ package com.clouway.oauth2.idtoken;
 
 
 import com.clouway.oauth2.jws.Pem;
-import com.clouway.oauth2.jws.RsaJwsSignature;
+import com.clouway.oauth2.jws.RsaJwsSigner;
 import com.clouway.oauth2.jwt.Jwt.ClaimSet;
 import com.clouway.oauth2.jwt.Jwt.Header;
 import com.google.common.io.BaseEncoding;
+
+import static com.clouway.oauth2.jws.RsaJwsSigner.sign;
 
 /**
  * @author Vasil Mitov <v.mitov.clouway@gmail.com>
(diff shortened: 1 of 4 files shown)


oauth2-server/src/main/java/com/clouway/oauth2/idtoken/IdTokenBuilder.java
```diff
@@ -40,8 +41,7 @@ public String build() {
     String claimsString = BaseEncoding.base64Url().omitPadding().encode(claims.toString().getBytes());
 
     String unsignedToken = String.format("%s.%s", headerString, claimsString);
-    RsaJwsSignature signature = new RsaJwsSignature(unsignedToken.getBytes());
 
-    return String.format("%s.%s", unsignedToken, signature.sign(unsignedToken.getBytes(), key));
+    return String.format("%s.%s", unsignedToken, sign(unsignedToken.getBytes(), key));
```
mgenov: RsaJwsSigner.sign() instead of just sign as it will indicate what signature is used.

oauth2-server/src/main/java/com/clouway/oauth2/jws/RsaJwsSigner.java
```diff
@@ -0,0 +1,41 @@
...
+  private static final SignatureAlgorithm rsaAlgorithm = SignatureAlgorithm.forName("RS256");
+
+  public static String sign(byte[] content, Pem.Block privateKey) {
+    Key prKey = parsePem(privateKey);
+    RsaSigner rsaSigner = new RsaSigner(rsaAlgorithm, prKey);
+    byte[] signedContent = rsaSigner.sign(content);
+    return tokenSignature(signedContent);
+  }
+
+  private static Key parsePem(Pem.Block privateKey) {
```
mgenov: It always return PrivateKey no matter that returned value is the abstraction. It's better to be concrete in this case.

mgenov (changes requested):

oauth2-server/src/main/java/com/clouway/oauth2/idtoken/IdTokenBuilder.java
VasilMitovClouway: Changed.

oauth2-server/src/main/java/com/clouway/oauth2/jws/RsaJwsSigner.java
VasilMitovClouway: Changed to PrivateKey

mgenov (approved):
LGTM !

Issue #46, opened 2017-04-11
oauth2: id_token is not a json formatted one


mgenov:
The generated id_token doesn't contain JSON value as ClaimSet is returning toString() representation of the object and not a json object.

VasilMitovClouway:
Was wondering about this issue - now that we have Jjwt library, we could just remove the existing id token builder and just use the librarys DefaultJwtBuilder.Tho this will make RsaJwsSigner( to be implemented in #51 ) obsolete. The other way is to just add toString() methods and properly construct fields to make a JSON.

mgenov:
Yes, jjwt need to be used for signing and generation of jwts.

mgenov:
Should be fixed with #52

VasilMitovClouway:
Fixed in  #52

PR #52, opened 2017-04-18
oauth2: idToken is now generated from Jjwt lib

VasilMitovClouway:
The Id token is now build by the DefaultJwtBuilder from the Jjwt library and is properly recognized and verified by RsaJwsSignature. Removed the RsaJwsSigner and IdTokenBuilder for they are no longer needed. 
Require discussion on: 
1) For future usage we discussed that certificate identifier is good to be added, so a cid field was added to the Header.
2) For convenience the whole Identity object is serialized and added to the id token. I gave it a thought and I think it's a good idea, but I am open for suggestions.

regarding #46


oauth2-server/src/main/java/com/clouway/oauth2/IssueNewTokenActivity.java
```diff
@@ -64,27 +72,27 @@ public Response execute(Client client, Request request, DateTime instant) {
...
+    Gson gson = new Gson();
+
+    String idToken = idTokenBuilder
+            .setHeaderParam("cid", certKey)//CertificateId - the ID of the certificate that the token was signed with.
+            .setIssuer(host)
+            .setAudience(client.id)
+            .setSubject(identity.id())
+            .setIssuedAt(instant.asDate())
+            .setExpiration(new Date(accessToken.expirationTimestamp()))
+            .claim("identity", gson.toJson(identity))
```
mgenov: This will make it different from the UserInfo and will enforce clients to use different parses for each. 

This is how UserInfo claims are build:
```java
    o.addProperty("id", identity.id());
    o.addProperty("name", identity.name());
    o.addProperty("email", identity.email());
    o.addProperty("given_name", identity.givenName());
    o.addProperty("family_name", identity.familyName());

    Map<String, Object> claims = identity.claims();

    for (String key : claims.keySet()) {
      Object value = claims.get(key);

      if (value instanceof String) {
        o.addProperty(key, (String) value);
      }
      if (value instanceof Number) {
        o.addProperty(key, (Number) value);
      }

      if (value instanceof Boolean) {
        o.addProperty(key, (Boolean) value);
      }
    }
```

mgenov (changes requested):

oauth2-server/src/test/java/DefaultJjwtBuilderTest.java
```diff
@@ -0,0 +1,60 @@
...
+
+import static org.junit.Assert.*;
+
+/**
+ * @author Vasil Mitov <v.mitov.clouway@gmail.com>
+ */
+public class DefaultJjwtBuilderTest {
+
+  @Test
+  public void generateAnIdTokenWithIdentityTest() throws Exception {
```
mgenov: Case when signature is not matching is not available.

mgenov (changes requested):

oauth2-server/src/main/java/com/clouway/oauth2/IssueNewTokenActivity.java
VasilMitovClouway: Added a method toJson to the Identity object and did some refactoring.

oauth2-server/src/test/java/DefaultJjwtBuilderTest.java
VasilMitovClouway: Added missing test case.

VasilMitovClouway:
Fixing failing check.

oauth2-server/src/main/java/com/clouway/oauth2/Identity.java
```diff
@@ -53,6 +56,32 @@ public String picture() {
     return claims;
   }
 
+  public JsonObject asJson(){
```
mgenov: I prefer to be moved out of this class instead of putting it in there, as it adds explicit JsonObject dependency in client code.  (accounts will be able to use it, which doesn't sounds reasonable)

UserInfoResponse response = UserInfoResponse.of(identity); //create user info from identity
response.asJson();


where the UserInfo will be packed as private class which will be visible only internally by the library.

oauth2-server/src/test/java/DefaultJjwtBuilderTest.java
```diff
@@ -0,0 +1,92 @@
...
+
+import static org.junit.Assert.*;
+
+/**
+ * @author Vasil Mitov <v.mitov.clouway@gmail.com>
+ */
+public class DefaultJjwtBuilderTest {
+
+  @Test
+  public void validIdTokenTest() throws Exception {
```
mgenov: tokenIsValid

oauth2-server/src/test/java/DefaultJjwtBuilderTest.java
```diff
@@ -0,0 +1,92 @@
...
+import java.security.spec.PKCS8EncodedKeySpec;
+import java.util.Collections;
+import java.util.Date;
+
+import static org.junit.Assert.*;
+
+/**
+ * @author Vasil Mitov <v.mitov.clouway@gmail.com>
+ */
+public class DefaultJjwtBuilderTest {
```
mgenov: I'm not sure which behaviour is actually tested here.

mgenov (changes requested):

oauth2-server/src/test/java/DefaultJjwtBuilderTest.java
VasilMitovClouway: Renamed

oauth2-server/src/test/java/DefaultJjwtBuilderTest.java
VasilMitovClouway: Renamed

oauth2-server/src/main/java/com/clouway/oauth2/IssueNewTokenActivity.java
```diff
@@ -64,27 +70,26 @@ public Response execute(Client client, Request request, DateTime instant) {
...
+            .setIssuer(host)
+            .setAudience(client.id)
+            .setSubject(identity.id())
+            .setIssuedAt(instant.asDate())
+            .setExpiration(new Date(accessToken.expirationTimestamp()))
+            .claim("identity", identityAsJson(identity).toString())
+            .signWith(SignatureAlgorithm.RS256, parsePem(key))
+            .compact();
 
     return new BearerTokenResponse(accessToken.value, accessToken.ttlSeconds(instant), response.refreshToken, idToken);
```
mgenov, line 93: The IssueNewTokenForClientTest doesn't verify that idToken is now returned from the response and it's a valid token.

mgenov (changes requested):

oauth2-server/src/main/java/com/clouway/oauth2/IssueNewTokenActivity.java
VasilMitovClouway: It is now tested for availability of id token in the body of the response.

oauth2-server/src/main/java/com/clouway/oauth2/IssueNewTokenActivity.java
```diff
@@ -64,27 +70,26 @@ public Response execute(Client client, Request request, DateTime instant) {
...
-                    .exp(accessToken.expirationTimestamp())
-                    .certId(certKey)
-                    .customClaims(identity.claims())
-                    .build())
-            .signWith(key)
-            .build();
+    String host = request.header("Host");
+    DefaultJwtBuilder idTokenBuilder = new DefaultJwtBuilder();
+
+    String idToken = idTokenBuilder
```
mgenov: You cab shorten it with the static builder method: Jwts.builder().....setHeaderParam();

This will remove the instantiation.

mgenov (changes requested):

oauth2-server/src/main/java/com/clouway/oauth2/IssueNewTokenActivity.java
VasilMitovClouway: Instantiation removed.

mgenov (approved):
LGTM !

PR #54, opened 2017-04-19
oauth2-server: auth code now is verified at specified time

mgenov:
The AuthorizationCode is now verified by using and the authorization time which allows on application to use short living authorization codes.

### DIFF
--- a/oauth2-example-app/src/main/java/com/clouway/oauth2/exampleapp/storage/InMemoryClientAuthorizationRepository.java
+++ b/oauth2-example-app/src/main/java/com/clouway/oauth2/exampleapp/storage/InMemoryClientAuthorizationRepository.java
@@ -1,5 +1,6 @@
 package com.clouway.oauth2.exampleapp.storage;
 
+import com.clouway.oauth2.DateTime;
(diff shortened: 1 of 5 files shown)


oauth2-example-app/src/main/java/com/clouway/oauth2/exampleapp/storage/InMemoryClientAuthorizationRepository.java
```diff
@@ -23,7 +24,7 @@ public InMemoryClientAuthorizationRepository(TokenGenerator tokenGenerator) {
   }
 
   @Override
-  public Optional<Authorization> findAuthorization(Client client, String authCode) {
+  public Optional<Authorization> findAuthorization(Client client, String authCode, DateTime instant) {
```
VasilMitovClouway, line 27: What will be the usage of the DateTime in our case ?

VasilMitovClouway:
Disregard my question, just saw the newly added documentation.

VasilMitovClouway (approved):

VasilMitovClouway:
LGTM!

PR #56, opened 2017-04-19
oauth2-server/api: make AuthorizedClientActivity private

mgenov:
AuthorizedClientActivity interface is now private and is not accessible from outside.


VasilMitovClouway (approved):

VasilMitovClouway:
LGTM!

PR #57, opened 2017-04-19
oauth2-server/jwt: return descriptions for responses

mgenov:
The JWT responses are now having descriptions which will improve remote debugging of client requests (they will receive better responses).


VasilMitovClouway (approved):

VasilMitovClouway:
LGTM!

PR #58, opened 2017-04-20
oauth2-example-app: find jwt key using iss

mgenov:
A lookup for the JWT key is now performed by the iss claim in the InMemoryCleintRepository.


VasilMitovClouway (approved):
LGTM!

PR #59, opened 2017-05-11
oauth2-server/id_token: minor changes satisfying specifications.

VasilMitovClouway:
Changes have been applied in IssueNewTokenActivity to the generated id_token claims. Now identity is provided as separate claims properly suiting specifications of https://tools.ietf.org/html/draft-ietf-oauth-json-web-token-32.
The chosen private claims are named "_private_claims" to try and avoid name collisions.

### DIFF
--- a/oauth2-server/src/main/java/com/clouway/oauth2/IssueNewTokenActivity.java
+++ b/oauth2-server/src/main/java/com/clouway/oauth2/IssueNewTokenActivity.java
@@ -10,7 +10,6 @@ import com.clouway.oauth2.token.BearerToken;
 import com.clouway.oauth2.token.GrantType;
 import com.clouway.oauth2.token.TokenResponse;
 import com.clouway.oauth2.token.Tokens;
-import com.google.gson.JsonObject;
 import io.jsonwebtoken.Jwts;
 import io.jsonwebtoken.SignatureAlgorithm;
 
@@ -20,7 +19,7 @@ import java.security.PrivateKey;
 import java.security.spec.InvalidKeySpecException;
 import java.security.spec.PKCS8EncodedKeySpec;
 import java.util.ArrayList;
-import java.util.Date;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Random;
@@ -57,14 +56,21 @@ class IssueNewTokenActivity implements AuthorizedClientActivity {
       Pem.Block key = certificates.get(certKey);
       String host = request.header("Host");
 
+      Map<String, Object> claims = new LinkedHashMap<String, Object>();
+      claims.put("iss", host);
+      claims.put("aud", client.id);
+      claims.put("sub", identity.id());
+      claims.put("iat", (instant.timestamp() / 1000));
+      claims.put("exp", accessToken.ttlSeconds(instant));
+      claims.put("name", identity.name());
+      claims.put("email", identity.email());
+      claims.put("given_name", identity.givenName());
+      claims.put("family_name", identity.familyName());
+      claims.putAll(identity.claims());
+
       idToken = Jwts.builder()
               .setHeaderParam("cid", certKey)//CertificateId - the ID of the certificate that the token was signed with.
-              .setIssuer(host)
-              .setAudience(client.id)
-              .setSubject(identity.id())
-              .setIssuedAt(instant.asDate())
-              .setExpiration(new Date(accessToken.expirationTimestamp()))
-              .claim("identity", identityAsJson(identity).toString())
+              .setClaims(claims)
               .signWith(SignatureAlgorithm.RS256, parsePem(key))
               .compact();
     }
@@ -86,33 +92,5 @@ class IssueNewTokenActivity implements AuthorizedClientActivity {
       e.printStackTrace();
     }
     return null;
-  }
-
-  private JsonObject identityAsJson(Identity identity) {
-    JsonObject o = new JsonObject();
-
-    o.addProperty("id", identity.id());
-    o.addProperty("name", identity.name());
-    o.addProperty("email", identity.email());
-    o.addProperty("given_name", identity.givenName());
-    o.addProperty("family_name", identity.familyName());
-
-    Map<String, Object> claims = identity.claims();
-
-    for (String key : claims.keySet()) {
-      Object value = claims.get(key);
-
-      if (value instanceof String) {
-        o.addProperty(key, (String) value);
-      }
-      if (value instanceof Number) {
-        o.addProperty(key, (Number) value);
-      }
-
-      if (value instanceof Boolean) {
-        o.addProperty(key, (Boolean) value);
-      }
-    }
-    return o;
   }
 }
--- a/oauth2-server/src/test/java/VerifyingJjwtTokensTest.java
+++ b/oauth2-server/src/test/java/VerifyingJjwtTokensTest.java
@@ -3,8 +3,8 @@ import com.clouway.oauth2.jws.Pem;
 import com.clouway.oauth2.jws.Pem.Block;
 import com.clouway.oauth2.jws.ReadPemFilesTest;
 import com.clouway.oauth2.jws.RsaJwsSignature;
+import com.google.common.collect.ImmutableMap;
 import com.google.common.io.BaseEncoding;
-import com.google.gson.JsonObject;
 import io.jsonwebtoken.Jwts;
 import io.jsonwebtoken.SignatureAlgorithm;
 import org.junit.Test;
@@ -14,8 +14,8 @@ import java.security.NoSuchAlgorithmException;
 import java.security.PrivateKey;
 import java.security.spec.InvalidKeySpecException;
 import java.security.spec.PKCS8EncodedKeySpec;
-import java.util.Collections;
 import java.util.Date;
+import java.util.Map;
 
 import static org.junit.Assert.*;
 
@@ -30,16 +30,24 @@ public class VerifyingJjwtTokensTest {
     Pem.Block block = pem.parse(ReadPemFilesTest.class.getResourceAsStream("secret.pem"));
 
     Identity identity = new Identity("123", "Pojo", "Foo", "Bar",
-            "example@email.com", "", Collections.EMPTY_MAP);
+            "example@email.com", "", ImmutableMap.<String, Object>of("customerId", "::customer Id::",
+            "customerName", "::customer name::", "isAdmin", false));
+
+    Map<String, Object> claims = ImmutableMap.<String, Object>builder()
+            .put("iss", "example.host")
+            .put("aud", "123")
+            .put("sub", identity.id())
+            .put("iat", new Date())
+            .put("exp", new Date())
+            .put("name", identity.name())
+            .put("email", identity.email())
+            .put("given_name", identity.givenName())
+            .put("family_name", identity.familyName())
+            .putAll(identity.claims()).build();
 
     String idToken = Jwts.builder()
             .setHeaderParam("cid", "certKeyIdentifier")
-            .setIssuer("host.app@example.com")
-            .setAudience("aud")
-            .setSubject(identity.id())
-            .setIssuedAt(new Date())
-            .setExpiration(new Date())
-            .claim("identity", new JsonObject().toString())
+            .setClaims(claims)
             .signWith(SignatureAlgorithm.RS256, getPrivateKey(block))
             .compact();
 
@@ -56,18 +64,16 @@ public class VerifyingJjwtTokensTest {
   public void tokenIsNotValidTest() throws Exception {
     Pem pem = new Pem();
     Pem.Block block = pem.parse(ReadPemFilesTest.class.getResourceAsStream("secret.pem"));
-
-    Identity identity = new Identity("123", "Pojos", "Foos", "Bars",
-            "examples@email.com", "", Collections.EMPTY_MAP);
+    Map<String, Object> claims = ImmutableMap.<String, Object>builder()
+            .put("iss", "example.host")
+            .put("aud", "123")
+            .put("sub", "::subject::")
+            .put("iat", new Date())
+            .put("exp", new Date()).build();
 
     String idToken = Jwts.builder()
             .setHeaderParam("cid", "certKeyIdentifiers")
-            .setIssuer("anotherHost.app@example.com")
-            .setAudience("aud")
-            .setSubject(identity.id())
-            .setIssuedAt(new Date())
-            .setExpiration(new Date())
-            .claim("identity", new JsonObject().toString())
+            .setClaims(claims)
             .signWith(SignatureAlgorithm.RS256, getPrivateKey(block))
             .compact();
 
--- a/oauth2-server/src/test/java/com/clouway/oauth2/IssueNewTokenForClientTest.java
+++ b/oauth2-server/src/test/java/com/clouway/oauth2/IssueNewTokenForClientTest.java
@@ -53,7 +53,7 @@ public class IssueNewTokenForClientTest {
       oneOf(keyStore).privateCertificates();
       will(returnValue(Collections.singletonMap("53r7IfiCaT3", generatePrivateKey())));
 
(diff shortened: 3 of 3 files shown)


oauth2-server/src/main/java/com/clouway/oauth2/IssueNewTokenActivity.java
```diff
@@ -64,7 +64,12 @@ public Response execute(Client client, Identity identity, Set<String> scopes, Re
               .setSubject(identity.id())
               .setIssuedAt(instant.asDate())
               .setExpiration(new Date(accessToken.expirationTimestamp()))
-              .claim("identity", identityAsJson(identity).toString())
+              .claim("id", identity.id())
```
mgenov: id could be changed to sub as it's standard claim. I know that in the past id was used, but sub is better.

oauth2-server/src/main/java/com/clouway/oauth2/IssueNewTokenActivity.java
```diff
@@ -64,7 +64,12 @@ public Response execute(Client client, Identity identity, Set<String> scopes, Re
...
+              .claim("name", identity.name())
+              .claim("email", identity.email())
+              .claim("given_name", identity.givenName())
+              .claim("family_name", identity.familyName())
+              .claim("_private_claims", privateClaims(identity).toString())
```
mgenov: Don't add them as additional property as the apps need to ensure that they are unique.  It is easier for parsing from the client side if they are at the same level as standard claims.

mgenov (changes requested):

oauth2-server/src/main/java/com/clouway/oauth2/IssueNewTokenActivity.java
VasilMitovClouway: As I understand right now the custom claims that we have are namespace and isAdmin. Should I just add them as _namespace and _isAdmin (adding the underscore just to be sure not to get those collisions) or should I try to hack my way around the cycle in Identity.claims() and then add them as the names that they have in that Map?

oauth2-server/src/main/java/com/clouway/oauth2/IssueNewTokenActivity.java
VasilMitovClouway: I will just remove it then, as it is defined in .setSubject()

oauth2-server/src/main/java/com/clouway/oauth2/IssueNewTokenActivity.java
mgenov: They could be hacked by the applications. Just use the same name that was stored as key in the map.

VasilMitovClouway: My mistake customerId and customerName are in the list as well.

VasilMitovClouway: No underscores, no nothing, im adding them as they are.

VasilMitovClouway:
Changes applied.

oauth2-server/src/main/java/com/clouway/oauth2/IssueNewTokenActivity.java
```diff
@@ -64,7 +64,14 @@ public Response execute(Client client, Identity identity, Set<String> scopes, Re
               .setSubject(identity.id())
               .setIssuedAt(instant.asDate())
               .setExpiration(new Date(accessToken.expirationTimestamp()))
-              .claim("identity", identityAsJson(identity).toString())
+              .claim("name", identity.name())
+              .claim("email", identity.email())
+              .claim("given_name", identity.givenName())
+              .claim("family_name", identity.familyName())
+              .claim("namespace", identity.claims().get("namespace"))
```
mgenov: You have to iterate over the map as these are custom claims and the library shouldn't know anything about them.

VasilMitovClouway: Changed. Now JJWT's 
```java 
.setClaims(Map<String,Object> claims)
 ````
 is used

File: oauth2-server/src/test/java/VerifyingJjwtTokensTest.java [hunk 4]
```diff
@@ -30,20 +33,27 @@ public void tokenIsValidTest() throws Exception {
     Pem.Block block = pem.parse(ReadPemFilesTest.class.getResourceAsStream("secret.pem"));
 
     Identity identity = new Identity("123", "Pojo", "Foo", "Bar",
-            "example@email.com", "", Collections.EMPTY_MAP);
+            "example@email.com", "", ImmutableMap.<String, Object>of("customerId", "::customer Id::",
+            "customerName", "::customer name::", "isAdmin", false));
+
+    Map<String, Object> claims = new LinkedHashMap<String, Object>();
```
mgenov: ImmutableMap.Builder could make things shorter.

oauth2-server/src/test/java/VerifyingJjwtTokensTest.java
```diff
@@ -56,18 +66,16 @@ public void tokenIsValidTest() throws Exception {
   public void tokenIsNotValidTest() throws Exception {
     Pem pem = new Pem();
     Pem.Block block = pem.parse(ReadPemFilesTest.class.getResourceAsStream("secret.pem"));
-
-    Identity identity = new Identity("123", "Pojos", "Foos", "Bars",
-            "examples@email.com", "", Collections.EMPTY_MAP);
+    Map<String, Object> claims = new LinkedHashMap<String, Object>();
```
mgenov: ImmutableMap.Builder could make things shorter.

oauth2-server/src/main/java/com/clouway/oauth2/IssueNewTokenActivity.java
```diff
@@ -57,14 +57,21 @@ public Response execute(Client client, Identity identity, Set<String> scopes, Re
       Pem.Block key = certificates.get(certKey);
       String host = request.header("Host");
 
+      Map<String, Object> claims = new LinkedHashMap<>();
+      claims.put("iss", host);
+      claims.put("aud", client.id);
+      claims.put("sub", identity.id());
+      claims.put("iat", instant.asDate());
+      claims.put("exp", new Date(accessToken.expirationTimestamp()));
```
mgenov: This is what docs says about exp: 
> exp
REQUIRED. Expiration time on or after which the ID Token MUST NOT be accepted for processing. The processing of this parameter requires that the current date/time MUST be before the expiration date/time listed in the value. Implementers MAY provide for some small leeway, usually no more than a few minutes, to account for clock skew. Its value is a JSON number representing the number of seconds from 1970-01-01T0:0:0Z as measured in UTC until the date/time. See RFC 3339 [RFC3339] for details regarding date/times in general and UTC in particula

**Its value is a JSON number representing the number of seconds from 1970-01-01T0:0:0Z as measured in UTC until the date/time**

oauth2-server/src/main/java/com/clouway/oauth2/IssueNewTokenActivity.java
```diff
@@ -57,14 +57,21 @@ public Response execute(Client client, Identity identity, Set<String> scopes, Re
...
+      claims.put("name", identity.name());
+      claims.put("email", identity.email());
+      claims.put("given_name", identity.givenName());
+      claims.put("family_name", identity.familyName());
+      claims.putAll(identity.claims());
```
mgenov, line 69: Have you checked that claims.putAll works with all primitive types such as boolean, long, integer, string  ?

There was a check with GSON but it need to be verified and here.

mgenov (changes requested):

oauth2-server/src/main/java/com/clouway/oauth2/IssueNewTokenActivity.java
VasilMitovClouway: Yes i tested it. Result after decode : 
{"iss":"example.host","aud":"123","sub":"123","iat":1494508639432,"exp":1494508639432,"name":"Pojo","email":"example@email.com","given_name":"Foo","family_name":"Bar","customerId":"::customer Id::","customerName":"::customer name::","isAdmin":false}

oauth2-server/src/main/java/com/clouway/oauth2/IssueNewTokenActivity.java
mgenov: The situation with the iat claim is same: 

> REQUIRED. Time at which the JWT was issued. Its value is a JSON number representing the number of seconds from 1970-01-01T0:0:0Z as measured in UTC until the date/time.

oauth2-server/src/main/java/com/clouway/oauth2/IssueNewTokenActivity.java
mgenov: looks good to me.

oauth2-server/src/main/java/com/clouway/oauth2/IssueNewTokenActivity.java
VasilMitovClouway: So it would be better to subtract a 60*1000 or 120*1000 from that time ?

VasilMitovClouway: In Java 7's Date class it says that :
A thin wrapper around a millisecond value that allows JDBC to identify this as an SQL DATE value. A milliseconds value represents the number of milliseconds that have passed since January 1, 1970 00:00:00.000 GMT. 

Isnt instant just that - a new Date() ?

VasilMitovClouway: So I need to add division by 100 to change the milliseconds to seconds, if I am understanding the problem correctly ?

mgenov: It has to be a long value, as documentation says number. Json serialization to Date() will produce string value unless serializer is producing long values.

mgenov: This JSON:

```javascript
{"iss":"example.host","aud":"123","sub":"123","iat":1494508639432,"exp":1494508639432,"name":"Pojo","email":"example@email.com","given_name":"Foo","family_name":"Bar","customerId":"::customer Id::","customerName":"::customer name::","isAdmin":false}
```
is fine as values are serialized as longs (I suppose jackson is doing it by default)

mgenov: But creation of new Date is overwhelming as it's one additional allocation.

mgenov: > So I need to add division by 100 to change the milliseconds to seconds, if I am understanding the problem correctly ?

If value is in milliseconds it has to be converted to seconds.

VasilMitovClouway: Is accessToken.expirationTimestamp() enought ?

mgenov: I suppose not, as it has to be in milliseconds. 

The BearerToken class has method about this: 

```java
 /**
   * Gets time to live in seconds of the current token.
   *
   * @param instant the instant time used for check
   * @return the time to live in seconds
   */
  public Long ttlSeconds(DateTime instant) {
    return (expiresAt.timestamp() - instant.timestamp()) / 1000;
  }
```

which calculates the time to live (exp) in seconds from the current time.

mgenov: This is a snippet from master:

```java
return new BearerTokenResponse(accessToken.value, accessToken.ttlSeconds(instant), response.refreshToken, idToken);
```

exp in token and exp in id_token should be same, so and the code should be same.

VasilMitovClouway: So i guess this 
```java 
      claims.put("iat", (instant.timestamp() /1000));
      claims.put("exp", accessToken.ttlSeconds(instant));
```
is good

mgenov: Yes, this should work as expected.

VasilMitovClouway: Implemented, as well as the previous comments.

oauth2-server/src/test/java/VerifyingJjwtTokensTest.java
mgenov: .

oauth2-server/src/test/java/VerifyingJjwtTokensTest.java
mgenov: .

oauth2-server/src/test/java/VerifyingJjwtTokensTest.java
VasilMitovClouway: I changed this.

oauth2-server/src/test/java/VerifyingJjwtTokensTest.java
```diff
@@ -30,16 +30,24 @@ public void tokenIsValidTest() throws Exception {
     Pem.Block block = pem.parse(ReadPemFilesTest.class.getResourceAsStream("secret.pem"));
 
     Identity identity = new Identity("123", "Pojo", "Foo", "Bar",
-            "example@email.com", "", Collections.EMPTY_MAP);
+            "example@email.com", "", ImmutableMap.<String, Object>of("customerId", "::customer Id::",
+            "customerName", "::customer name::", "isAdmin", false));
+
+    Map claims = ImmutableMap.builder()
```
mgenov:
```java
Map<String, Object> claims
```

oauth2-server/src/test/java/VerifyingJjwtTokensTest.java
```diff
@@ -56,18 +64,16 @@ public void tokenIsValidTest() throws Exception {
   public void tokenIsNotValidTest() throws Exception {
     Pem pem = new Pem();
     Pem.Block block = pem.parse(ReadPemFilesTest.class.getResourceAsStream("secret.pem"));
-
-    Identity identity = new Identity("123", "Pojos", "Foos", "Bars",
-            "examples@email.com", "", Collections.EMPTY_MAP);
+    Map claims = ImmutableMap.builder()
```
mgenov:
```java
Map<String, Object> claims
```

mgenov (approved):
LGTM !

mgenov:
CI is failing: 
happyPath(com.clouway.oauth2.IssueNewTokenForClientTest)  Time elapsed: 0.099 sec  <<< ERROR!
java.lang.NullPointerException: null value in entry: iss=null
	at com.google.common.collect.CollectPreconditions.checkEntryNotNull(CollectPreconditions.java:33)
	at com.google.common.collect.ImmutableMap.entryOf(ImmutableMap.java:135)
	at com.google.common.collect.ImmutableMap$Builder.put(ImmutableMap.java:206)
	at com.clouway.oauth2.IssueNewTokenActivity.execute(IssueNewTokenActivity.java:59)
	at com.clouway.oauth2.IssueNewTokenForClientTest.happyPath(IssueNewTokenForClientTest.java:60)

VasilMitovClouway:
Strange, I haven't changed anything on that.

mgenov:
The ImmutableMap of Guava is not accepting null values whether in some test probably is used one. It's questionable here where the client apps which are using this library should be enforced to skip nulls or they will be allowed. If we are going to allow this, then immutable map usage is not possible. It may be used in tests but not in the production code.

VasilMitovClouway:
Im not sure what 
```java
client.description()
```
 is. Could it be used when 
```java
req.param("Host") == null
```

VasilMitovClouway:
This wouldn't fix the failing test tho.

VasilMitovClouway:
Map changed back to LinkedHashMap. All tests passed.

PR #61, opened 2017-05-12
oauth2-server/id_token: exp and iat change

VasilMitovClouway:
Another change regarding iat and exp. After looking around in the JJWT library I saw that the methods for setting iat and exp requier a Date format. I think it is better to let the library do the transformation and add it through it's methods setIssuedAt and setExparation. I have tested around with it and the format that it creates is in proper seconds.

### DIFF
--- a/oauth2-server/src/main/java/com/clouway/oauth2/IssueNewTokenActivity.java
+++ b/oauth2-server/src/main/java/com/clouway/oauth2/IssueNewTokenActivity.java
@@ -19,6 +19,7 @@ import java.security.PrivateKey;
 import java.security.spec.InvalidKeySpecException;
 import java.security.spec.PKCS8EncodedKeySpec;
 import java.util.ArrayList;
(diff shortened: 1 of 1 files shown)


VasilMitovClouway:
I think the main problem was that 
```java 
accessToken.ttlSeconds(instant)
``` 
returns the time to live in seconds and not seconds passed since January first 1970 (the required one)

mgenov:
Yeah it returns the seconds elapsed till the expiration time from the current time.

VasilMitovClouway:
Ive done some tests, we can skip the new Date() but instead of the accessToken.ttlSeconds(instant) I can use the previous accessToken.exparationTimestamp/1000.

VasilMitovClouway:
But i think the one with date is more secure as it is the one that the library uses.

mgenov:
If library provides expiration time to be passed as Date then it's the right way for doing it.

PR #62, opened 2017-06-05
oauth2-server: support scopes in token responses

mgenov:
The scopes which are allowed for resource owners are now returned in Token responses either when token is issued for the first time or when a refresh token is used for issuing of a new access token.


VasilMitovClouway (approved):
LGTM

PR #67, opened 2017-06-26
oauth2: allow authnetication from public clients

### LINKED ISSUE #65: oauth2: allow authentication from public clients
The oauth2-server library should allow authentication from public clients. Such clients in most cases are native apps (Android, iOS). 

The oauth2-server is currently verifying Client using Basic Authorization header but it need to be changed to verify whether client is a public one or not and the basic authorization header should not be required anymore. 

>  native application
      A native application is a public client installed and executed on
      the device used by the resource owner.  Protocol data and
      credentials are accessible to the resource owner.  It is assumed
      that any client authentication credentials included in the
      application can be extracted.  On the other hand, dynamically
      issued credentials such as access tokens or refresh tokens can
      receive an acceptable level of protection.  At a minimum, these
      credentials are protected from hostile servers with which the
      application may interact.  On some platforms, these credentials
      might be protected from other applications residing on the same
      device.

Spec: [RFC6749](https://tools.ietf.org/html/rfc6749)

mgenov:
With this change the oauth2-server library now allows authentication of public clients such as native apps (Android, iOS).

The native apps are not secure to keep the shared secret of the key and this is why in this situations a public key need to be used.

Fixes #65


VasilMitovClouway (approved):
LGTM!

PR #70, opened 2017-07-03
oauth2-server: improve key api design for id_token and public certs api

mgenov:
The design of the PublicKeys was improved by introducing a new object IdentityKeyPair which to represent a pair of keys (public and private) and an identity which is used to mark the id_token by keeping it in the header of the token.

In the past a pair of string <-> string was used, which leads to complications in the client apps and this is why the new object was introduced.

As ClientKeyStore is used only to identify service accounts JWT's signature it's renamed to JwtKeyStore which is more concrete and more understandable. The privateCertificates method is removed from this interface as IdentityKeyPair now represents Private and Publick keys and looks that it doesn't belong to this interface as JWT keys are different from the one used for the id_token.


VasilMitovClouway (approved):
LGTM!

PR #71, opened 2017-07-06
oauth2-server: provide scopes for the resource owner

mlesikov:
now the identityId and the scopes of the resource owner are
provided to the autorization.

### DIFF
--- a/oauth2-server/pom.xml
+++ b/oauth2-server/pom.xml
@@ -4,7 +4,7 @@
   <parent>
     <groupId>com.clouway.security</groupId>
     <artifactId>parent</artifactId>
-    <version>1.0-RC30-SNAPSHOT</version>
+    <version>1.0-RC31-SNAPSHOT</version>
   </parent>
   <groupId>com.clouway.security</groupId>
   <artifactId>oauth2-server</artifactId>
--- a/oauth2-server/src/main/java/com/clouway/oauth2/ClientAuthorizationActivity.java
+++ b/oauth2-server/src/main/java/com/clouway/oauth2/ClientAuthorizationActivity.java
@@ -27,7 +27,7 @@ class ClientAuthorizationActivity implements IdentityActivity {
   }
 
   @Override
-  public Response execute(String identityId, Request request) {
+  public Response execute(ResourceOwnerIdentity identity, Request request) {
     String responseType = request.param("response_type");
     String clientId = request.param("client_id");
     String requestedUrl = request.param("redirect_uri");
@@ -49,7 +49,7 @@ class ClientAuthorizationActivity implements IdentityActivity {
     }
 
     Set<String> scopes = Sets.newTreeSet(Splitter.on(" ").omitEmptyStrings().split(scope));
-    Optional<Authorization> possibleAuthorizationResponse = clientAuthorizationRepository.authorize(client, identityId, scopes, responseType);
+    Optional<Authorization> possibleAuthorizationResponse = clientAuthorizationRepository.authorize(client, identity.idenityId, identity.scopes, responseType);
 
     // RFC-6749 - Section: 4.2.2.1
     // The authorization server redirects the user-agent by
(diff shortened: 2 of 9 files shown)


mlesikov:
@mgenov  can take a look, please

oauth2-server/pom.xml
```diff
@@ -4,7 +4,7 @@
   <parent>
     <groupId>com.clouway.security</groupId>
     <artifactId>parent</artifactId>
-    <version>1.0-RC30-SNAPSHOT</version>
+    <version>1.0-RC31-SNAPSHOT</version>
```
mgenov, line 7: This should not be changed as release plugin is making it automatically.

oauth2-server/src/test/java/VerifyingJjwtTokensTest.java
```diff
@@ -61,6 +68,24 @@ public void tokenIsValidTest() throws Exception {
...
+    claims.put("id", 123456L);
+    String token = Jwts.builder()
+            .setHeaderParam("cid", "bla")//CertificateId - the ID of the certificate that the token was signed with.
+            .setIssuedAt(new Date())
+            .setClaims(claims)
+            .compact();
+
+    Jwt jwt = new DefaultJwtParser().parse(token);
+    DefaultClaims dc = (DefaultClaims) jwt.getBody();
+//    assertThat(dc.get("id", Long.class), is(123456L));
```
mgenov: get("id", String.class) should work as expected. Why it's commented ?

oauth2-server/src/main/java/com/clouway/oauth2/ClientAuthorizationActivity.java
```diff
@@ -49,7 +49,7 @@ public Response execute(String identityId, Request request) {
     }
 
     Set<String> scopes = Sets.newTreeSet(Splitter.on(" ").omitEmptyStrings().split(scope));
-    Optional<Authorization> possibleAuthorizationResponse = clientAuthorizationRepository.authorize(client, identityId, scopes, responseType);
+    Optional<Authorization> possibleAuthorizationResponse = clientAuthorizationRepository.authorize(client, identity.idenityId, identity.scopes, responseType);
```
mgenov, line 52: The passed scopes should be from the request, not from the identity as the Client app is requesting them. With this change it doesn't matter what is requesting the client.

mgenov (changes requested):

oauth2-server/src/main/java/com/clouway/oauth2/ClientAuthorizationActivity.java
mgenov: https://tools.ietf.org/html/rfc6749#section-4.2.1

oauth2-server/src/test/java/VerifyingJjwtTokensTest.java
mlesikov: ооо that is my previos test that I should not include in the PR, will be deleted  -no chenges over this classes

oauth2-server/src/main/java/com/clouway/oauth2/ClientAuthorizationActivity.java
mgenov: >  The authorization server MAY fully or partially ignore the scope
   requested by the client, based on the authorization server policy or
   the resource owner's instructions.  If the issued access token scope
   is different from the one requested by the client, the authorization
   server MUST include the "scope" response parameter to inform the
   client of the actual scope granted.

But as this is library, it should be decided by the client code which to be used and how, not by the library. The library should pass what was requested.

mlesikov: we have the flexibility to choose which scopes to return
we have the request(requested scopes in it) and we have the SID and client id - so we can find the exact user with the exact scopes for the application(clientId) and we can do what ever we wont with them in the omplementation in the accounts app.

mgenov: Yeah but I didn't get the reason why the design need to be changed ?

mlesikov: At this time there is no place where the user scopes(for bss-app) to be provided to the token, except the authorization request. The idea was these scopes to be stored for each user in the accouns app and when and provided to the api-gateway on when the accounts provides accessToken. 

With this change I move the decision which scopes to be provided out of the library. In that ResourceOwnerIdentityFInder.find is provided the request which should contains the requested scopes, clinetID and SID, so we have all the information that we need to decide and to check the requested scopes

mgenov: Please don't use real project names, just give information with interfaces and classes from this one.

mlesikov:
outdated

PR #72, opened 2017-07-06
oauth2-server: provide custom request params to the access token

mlesikov:
Handle index param of the request and pass it to the identityFinder.
This index can be used for selecting accounts of specific namespace

### DIFF
--- a/oauth2-server/src/main/java/com/clouway/oauth2/AuthCodeAuthorization.java
+++ b/oauth2-server/src/main/java/com/clouway/oauth2/AuthCodeAuthorization.java
@@ -7,7 +7,10 @@ import com.clouway.oauth2.authorization.ClientAuthorizationRepository;
 import com.clouway.oauth2.client.Client;
 import com.clouway.oauth2.token.GrantType;
 import com.clouway.oauth2.user.IdentityFinder;
+import com.clouway.oauth2.util.Params;
 import com.google.common.base.Optional;
+
+import java.util.Map;
 
 /**
  * @author Miroslav Genov (miroslav.genov@clouway.com)
@@ -18,7 +21,7 @@ class AuthCodeAuthorization implements ClientActivity {
   private final IdentityFinder identityFinder;
   private final AuthorizedClientActivity authorizedClientActivity;
 
-  AuthCodeAuthorization(ClientAuthorizationRepository clientAuthorizationRepository, IdentityFinder identityFinder,  AuthorizedClientActivity authorizedClientActivity) {
+  AuthCodeAuthorization(ClientAuthorizationRepository clientAuthorizationRepository, IdentityFinder identityFinder, AuthorizedClientActivity authorizedClientActivity) {
     this.clientAuthorizationRepository = clientAuthorizationRepository;
     this.identityFinder = identityFinder;
     this.authorizedClientActivity = authorizedClientActivity;
@@ -36,12 +39,14 @@ class AuthCodeAuthorization implements ClientActivity {
 
     Authorization authorization = possibleAuthorization.get();
 
-    Optional<Identity> possibleIdentity = identityFinder.findIdentity(authorization.identityId, GrantType.AUTHORIZATION_CODE, instant);
+    Map<String, String> params  = new Params().parse(request, "code");
+
+    Optional<Identity> possibleIdentity = identityFinder.findIdentity(authorization.identityId, GrantType.AUTHORIZATION_CODE, instant, params);
     if (!possibleIdentity.isPresent()) {
       return OAuthError.invalidGrant("identity was not found");
     }
-    
+
     Identity identity = possibleIdentity.get();
-    return authorizedClientActivity.execute(client, identity, authorization.scopes, request, instant);
+    return authorizedClientActivity.execute(client, identity, authorization.scopes, request, instant, params);
   }
 }
--- a/oauth2-server/src/main/java/com/clouway/oauth2/AuthorizedClientActivity.java
+++ b/oauth2-server/src/main/java/com/clouway/oauth2/AuthorizedClientActivity.java
@@ -4,6 +4,7 @@ import com.clouway.friendlyserve.Request;
 import com.clouway.friendlyserve.Response;
 import com.clouway.oauth2.client.Client;
 
+import java.util.Map;
 import java.util.Set;
 
 /**
@@ -19,7 +20,8 @@ interface AuthorizedClientActivity {
    * @param client  the client of which request will be handled
    * @param request the request
    * @param instant the time of which client was requested access
+   * @param params
    * @return the response
    */
-  Response execute(Client client, Identity identity, Set<String> scopes, Request request, DateTime instant);
+  Response execute(Client client, Identity identity, Set<String> scopes, Request request, DateTime instant, Map<String, String> params);
 }
--- a/oauth2-server/src/main/java/com/clouway/oauth2/token/BearerToken.java
+++ b/oauth2-server/src/main/java/com/clouway/oauth2/token/BearerToken.java
@@ -4,6 +4,7 @@ import com.clouway.oauth2.DateTime;
 import com.google.common.base.Objects;
 
 import java.io.Serializable;
+import java.util.Map;
 import java.util.Set;
 
 /**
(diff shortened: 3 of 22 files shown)


oauth2-server/src/main/java/com/clouway/oauth2/AuthorizedClientActivity.java
```diff
@@ -21,5 +21,5 @@
    * @param instant the time of which client was requested access
    * @return the response
    */
-  Response execute(Client client, Identity identity, Set<String> scopes, Request request, DateTime instant);
+  Response execute(Client client, Identity identity, Set<String> scopes, Request request, DateTime instant, int index);
```
mgenov: You can pass all params with something more common like: ```Map<String, String>``` instead of just index. 

Whether a list of allowed parameters could be passed as parameter to ClientAuthorizationActivity to prevent re-passing of any parameter. 

A default set of code and state could be passed if clients which are not using the library don't specify anything.

mgenov (changes requested):

oauth2-server/src/main/java/com/clouway/oauth2/token/BearerToken.java
```diff
@@ -15,21 +15,23 @@
...
   public final Set<String> scopes;
   private final DateTime expiresAt;
 
   public BearerToken() {
-    this(null, null, null, null, null, null, null);
+    this(null, null, null, null, null, null, null, 0);
   }
 
-  public BearerToken(String value, GrantType grantType, String identityId, String clientId, String email, Set<String> scopes, DateTime expiresAt) {
+  public BearerToken(String value, GrantType grantType, String identityId, String clientId, String email, Set<String> scopes, DateTime expiresAt, int index) {
```
qnislavnachev: why this index is in bearer token ? I think that index is never used from  bearer token.

qnislavnachev (changes requested):

oauth2-server/src/main/java/com/clouway/oauth2/token/BearerToken.java
qnislavnachev: I see now, we call findIdentity() when return currentUser and there index will be passed from token.

oauth2-server/src/main/java/com/clouway/oauth2/AuthCodeAuthorization.java
```diff
@@ -18,14 +21,17 @@
...
+  AuthCodeAuthorization(ClientAuthorizationRepository clientAuthorizationRepository, IdentityFinder identityFinder, AuthorizedClientActivity authorizedClientActivity) {
     this.clientAuthorizationRepository = clientAuthorizationRepository;
     this.identityFinder = identityFinder;
     this.authorizedClientActivity = authorizedClientActivity;
   }
 
   @Override
   public Response execute(Client client, Request request, DateTime instant) {
+
+    RequestExt req = (RequestExt) request;
```
qnislavnachev: As I can see RequestExt object is not used

oauth2-server/src/main/java/com/clouway/oauth2/AuthorizedClientActivity.java
mlesikov: changes applied

oauth2-server/src/main/java/com/clouway/oauth2/CustomParamsParser.java
```diff
@@ -0,0 +1,33 @@
...
+  public Map<String, String> parse(Request request, String... exclusions) {
+
+    RequestExt req = (RequestExt) request;
+
+    List<String> exclusionsList = Arrays.asList(exclusions);
+
+    Map<String, String> params = Maps.newHashMap();
+
+    for (String key : req.params().keySet()) {
+      String value = request.param(key) == null ? "" : request.param(key);
```
qnislavnachev: this check is unneeded request.params() will return only params that are available

oauth2-server/src/main/java/com/clouway/oauth2/authorization/ClientAuthorizationRepository.java
```diff
@@ -31,6 +31,7 @@
    * @param identityId the identity to be authorized
    * @param scopes scopes for the authorization code
    * @param responseType the type of the response to be returned
+   * @param index
```
qnislavnachev: this should be removed from documentation of the method

qnislavnachev (changes requested):

oauth2-server/src/main/java/com/clouway/oauth2/AuthCodeAuthorization.java
mlesikov: for (String key : req.params().keySet()) {

I need to get the prams()

oauth2-server/src/main/java/com/clouway/oauth2/authorization/ClientAuthorizationRepository.java
mlesikov: done

oauth2-server/src/main/java/com/clouway/oauth2/AuthCodeAuthorization.java
mlesikov: deleted

qnislavnachev (approved):
LGTM

oauth2-server/src/test/java/com/clouway/oauth2/util/ParamRequest.java
```diff
@@ -0,0 +1,85 @@
...
+import java.io.ByteArrayInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * @author Mihail Lesikov (mlesikov@gmail.com)
+ */
+public class ParamRequest implements RequestExt {
```
mgenov: FakeRequest class of fserve already should provide this.

oauth2-server/src/main/java/com/clouway/oauth2/AuthCodeAuthorization.java
```diff
@@ -36,12 +38,14 @@ public Response execute(Client client, Request request, DateTime instant) {
 
     Authorization authorization = possibleAuthorization.get();
 
-    Optional<Identity> possibleIdentity = identityFinder.findIdentity(authorization.identityId, GrantType.AUTHORIZATION_CODE, instant);
+    Map<String, String> params  = new CustomParamsParser().parse(request, "code");
```
mgenov: This doesn't look readable as custom is too common. 

Maybe something like: 
```java
Map<String, String> params = Params.parse(request, excluding("code")) 
```
reads better.

mgenov (changes requested):

oauth2-server/src/test/java/com/clouway/oauth2/util/ParamRequest.java
mlesikov: no, it doe not implements RequestExt, it implement just Request and there was a problem with the type cast  - coud not typecas ParamRequest to RequestExt.
I saw it and would like to use it, but the fserve is in the maven repo  and I have to call you to update and... 

I will add the change in the fserve and we can do the magic

mgenov: https://github.com/clouway/fserve/blob/master/testing/src/main/java/com/clouway/friendlyserve/testing/FakeRequest.java#L14

mgenov: As I can see it implements RequestExt already.

oauth2-server/src/main/java/com/clouway/oauth2/AuthCodeAuthorization.java
mlesikov: I agree, applied

now looks like this 
image

oauth2-server/src/test/java/com/clouway/oauth2/util/ParamRequest.java
mlesikov: all the ussage of ParamRequest was migrated and now only FakeRequest is used

mgenov (approved):
LGTM !

mgenov:
Latest changes are released in 1.0: https://oss.sonatype.org/content/repositories/releases/com/clouway/security/oauth2-server/1.0/

PR #74, opened 2017-08-14
oauth2-server: fixed rule patterns

KrasimirRaikov:
fixed rule patterns of jackson in jarjar plugin

### DIFF
--- a/oauth2-server/pom.xml
+++ b/oauth2-server/pom.xml
@@ -134,17 +134,17 @@
 
                 <rule>
                   <pattern>com.fasterxml.jackson.core.**</pattern>
-                  <result>com.clouway.security.internal.jackson-core.@1</result>
+                  <result>com.clouway.security.internal.jackson.core.@1</result>
                 </rule>
 
                 <rule>
-                  <pattern>com.fasterxml.jackson.core.**</pattern>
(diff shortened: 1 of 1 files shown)


KrasimirRaikov:
@mgenov the rule patterns were mistaken and were making trouble when trying to use jackson library.
I fixed them, but then it turned out that the jarjar wasn't needed. If you need it for some reason I can just fix the patterns

mgenov:
It's required by the jjwt library as it's transient dependent. 

e.g  oauth2-server -> jjwt -> jackson 

You can verify it by try to use the jjwt functionality.

mgenov:
By jarjaring jackson we are not forcing client apps which are using the library to include and whole jackson lib.

KrasimirRaikov:
I have returned the jarjar plugin and have fixed the rule patterns, also the result packaging

mgenov (approved):
LGTM !

KrasimirRaikov:
@mgenov could you update the oauth2 in the sonatype repository, because I have changes which depend on the jackson library in accounts, and right now their is a problem with the dependency

mgenov:
v1.0.1 is released and is available in the sonatype repository: https://oss.sonatype.org/content/repositories/releases/com/clouway/security/oauth2-server/1.0.1/

Issue #66, opened 2017-06-25
oauth2: proof key for code exchange


mgenov:
Native apps are vulnerable to attacks caused by interception of authorization code. They may hook up the authorization code and to exchange it for access token which to grant an access to user resources. 

The protection mechanism is described at: 
https://tools.ietf.org/html/rfc7636

mgenov:
https://medium.com/@ageitgey/everything-you-need-to-know-about-implementing-ios-and-android-mobile-deep-linking-f4348b265b49

Implementing Android App Links with Verification section shows an example for verification mechanism which could be used.

VasilMitovClouway:
A quick question: should we keep a backwards compatibility for a normal OAuth2 protocol without this extension ?

VasilMitovClouway:
For instance if the authorization request does not contain a code_verifyer a normal flow is done

mgenov:
Yes, normal flow should be executed if code_verifier is not passed by the client apps. 

oauth2-server should stick to standard oauth2 spec  but should support and additional security extensions which are defined latter.

VasilMitovClouway:
Another question: Will we encounter a problem using Sha 256 algorithm for the flow? Is there anything i should consider? I am thinking of using the "s256" method instead of the "plain"  ofc.

mgenov:
It would be better both to be supported. 

s256 is better option as start

PR #80, opened 2017-12-06
oauth2-server: implemented PKCE in authorization code flow

VasilMitovClouway:
A Proof Key for Code Exchange verification has been added to the authorization flow. 
The flow was previously vulnerable to a Man In The Middle attack for authorization code exchange when the code was returned to the Client. 
More on the matter described here: https://tools.ietf.org/html/rfc7636

Now the client can choose to add a code_challenge value alongside with code_challenge_method to the request for authorization code. If provided that CodeChallenge object needs to be stored with the provided authorization code. ( A ) 

When the client sends the authorization code in the request for a new token he sends a code_verifier 
to be compared with the previously provided code_challenge. ( C ) 

image

code_verifier - code_challenge pairs are generated on the client side as described in the rfc7636. 

If no code_challenge is provided then an empty CodeChallenge object should be stored with the authorization code. 

I could do a cleanup of the Authorization constructor and ClientAuthorizationRepository's method authorize in this PR if these changes are accepted.

Fixes: #66


oauth2-server/src/main/java/com/clouway/oauth2/AuthCodeAuthorization.java
```diff
@@ -20,33 +22,38 @@
...
   private final IdentityFinder identityFinder;
   private final AuthorizedClientActivity authorizedClientActivity;
+  private final CodeVerifier codeVerifier;
 
-  AuthCodeAuthorization(ClientAuthorizationRepository clientAuthorizationRepository, IdentityFinder identityFinder, AuthorizedClientActivity authorizedClientActivity) {
+  AuthCodeAuthorization(ClientAuthorizationRepository clientAuthorizationRepository, IdentityFinder identityFinder, AuthorizedClientActivity authorizedClientActivity, CodeVerifier codeVerifier) {
     this.clientAuthorizationRepository = clientAuthorizationRepository;
     this.identityFinder = identityFinder;
     this.authorizedClientActivity = authorizedClientActivity;
+    this.codeVerifier = codeVerifier;
```
mgenov: You can move codeVerification check one level up to the callstack, e.g in another implementation of ClientActivity. This will remove this new dependency here and will make tests clear for this part. 

```java
// e.g code exchange guards calling of AuthCodeAuthorization 
new CodeExchangeVerificationFlow(codeVerifier, new AuthCodeAuthorization()) 
```

oauth2-server/src/main/java/com/clouway/oauth2/codechallenge/AuthorizationCodeVerifier.java
```diff
@@ -0,0 +1,53 @@
...
+  @Override
+  public boolean verify(CodeChallenge codeChallenge, String codeVerifier) {
+    try {
+      //no code verifier was provided and no code challenge was saved so this is a normal OAuth2 code flow without PKCE
+      if (isNullOrEmpty(codeVerifier) && isNullOrEmpty(codeChallenge)) {
+        return true;
+      }
+
+      if (isNullOrEmpty(codeVerifier) && !isNullOrEmpty(codeChallenge)) {
+        //there is a saved code challenge but provided codeVerifier is null or empty.
```
mgenov: comment should be moved before the if statement as in previous block

oauth2-server/src/main/java/com/clouway/oauth2/codechallenge/AuthorizationCodeVerifier.java
```diff
@@ -0,0 +1,53 @@
...
+        String transformedCodeVerifier = BaseEncoding.base64Url().encode(hashed);
+        return transformedCodeVerifier.equals(codeChallenge.transformedCodeChallenge);
+      }
+      return false;
+    } catch (NoSuchAlgorithmException e) {
+      return false;
+    }
+  }
+
+  private boolean isNullOrEmpty(String string) {
```
mgenov: guava has method isNullOrEmpty.

oauth2-server/src/main/java/com/clouway/oauth2/authorization/ClientAuthorizationRepository.java
```diff
@@ -2,38 +2,51 @@
...
+   * Authorize authorizes clients with an included codeChallenge for better security
+   *
+   * @param client        the client which is requesting authorization
+   * @param identityId    the identity to be authorized
+   * @param scopes        scopes for the authorization code
+   * @param responseType  the type of the response to be returned
+   * @param codeChallenge the code challenge sent by the client
+   * @return the authorization or absent value if the identity was not authorized
+   */
+  Optional<Authorization> authorize(Client client, String identityId, Set<String> scopes, String responseType, CodeChallenge codeChallenge);
```
mgenov: This method becomes good candidate for Dependency Inversion: 

```java
class CodeChallengeAuthorization implements ClientAuthorization {

  public CodeChallengeAuthorization(CodeChallenge codeChallenge) {
    // ...
  }
  public Optional<Authorization> authenticate(Client client, String identity, Set<String> scopes, String responseType){
    // ...     
  }
}

class BasicAuthorization implements ClientAuthorization {
    public Optional<Authorization> authenticate(Client client, String identity, Set<String> scopes, String responseType){
    // ...
  }
}
```

e.g the moving parts between implementations

oauth2-server/src/main/java/com/clouway/oauth2/codechallenge/AuthorizationCodeVerifier.java
```diff
@@ -0,0 +1,53 @@
...
+        return true;
+      }
+
+      if (isNullOrEmpty(codeVerifier) && !isNullOrEmpty(codeChallenge)) {
+        //there is a saved code challenge but provided codeVerifier is null or empty.
+        return false;
+      }
+
+      if (!isNullOrEmpty(codeVerifier) && isNullOrEmpty(codeChallenge)) {
+        //codeVerifier was provided but no codeChallenge was saved.
```
mgenov: .

oauth2-server/src/main/java/com/clouway/oauth2/authorization/ClientAuthorizationRepository.java
```diff
@@ -2,38 +2,51 @@
...
-   * @param client the client for which authorization was issued
+   * @param client   the client for which authorization was issued
    * @param authCode the code of the authorization
-   * @param instant the time on which check is performed
+   * @param instant  the time on which check is performed
    * @return the authorization or absent value
    */
   Optional<Authorization> findAuthorization(Client client, String authCode, DateTime instant);
-
+//If we decide to use the new authorize method for all requests we could remove this as it is only used in the example app
```
mgenov: I'm not sure that I understand the goal of these comments. Are they going to be removed before PR is merged or there role is to be in the code  ?

mgenov (changes requested):

oauth2-server/src/main/java/com/clouway/oauth2/authorization/ClientAuthorizationRepository.java
VasilMitovClouway: This will be removed. Forgot to write it in this one as well.

oauth2-server/src/main/java/com/clouway/oauth2/codechallenge/AuthorizationCodeVerifier.java
VasilMitovClouway: Forgot about Strings.isNullOrEmpty. Implemented and removed method.

oauth2-server/src/main/java/com/clouway/oauth2/codechallenge/AuthorizationCodeVerifier.java
VasilMitovClouway: Moved.

oauth2-server/src/main/java/com/clouway/oauth2/codechallenge/AuthorizationCodeVerifier.java
VasilMitovClouway: Moved.

oauth2-server/src/main/java/com/clouway/oauth2/AuthCodeAuthorization.java
VasilMitovClouway: Wouldn't this chain of responsibility duplicate the roundtrip to the database for Authorization object ? We will check if the codeChallenge stored with the Authorization matches then in the AuthCodeAuthorization we extract the authorization again. Thinking for a solution.

VasilMitovClouway: Or I guess we can do the authorization first and if the verification passes then we will return it from this new class.

mgenov: It will perform just existence and syntax check. 

The first 2 if statements which are turning OAuth2Error in case of match.

mgenov: Yeah, it may force you to add a new interface which has execute method that accepts Authorization.

oauth2-server/src/main/java/com/clouway/oauth2/authorization/ClientAuthorizationRepository.java
VasilMitovClouway: Could you elaborate more on the matter? I don't think i am understanding the idea correctly.

mgenov: It's not a good idea methods to be designed to accepts parameters which will be used only in some cases.

VasilMitovClouway: We want to support both ways of authorization. It looks strange but i think that in the case when an authorization code is generated a blank CodeChallenge to be sent to the repository. A possible solution to avoid this would be to separate the CodeChallenge and not save it with the Authorization object in the ClientAuthorizationRepository. Still i think they should be kept together as they are tightly coupled.

mgenov: Introduction of 2 implementation will force re-design of the authorization store. 

It maybe hidden by introducing of client authorization request. It will not look explicit in the implementation 

```java
class ClientAuthorizationRequest {
   Client client;
   String  identity;
   String responseType;   
   CodeChallenge challenge;  //nullable type 
}

// definition: 
Optional<Authorization> authenticate(ClientAuthorizationRequest request)
```

mgenov: btw, naming here is misleading: 
method says that it's authentication but returns authorization. I suppose method should be named authorize instead.

VasilMitovClouway: Originally the method is called authorize yeah. This looks better as well. We will have to do a change in the implementation of ClientAuthorizationRepository anyway.

mgenov: Yes, challenge need to be persisted and used later for the code exchange call.

oauth2-server/src/main/java/com/clouway/oauth2/authorization/AuthorizationRequest.java
```diff
@@ -1,38 +1,45 @@
...
  * @author Ivan Stefanov <ivan.stefanov@clouway.com>
  */
 public class AuthorizationRequest {
   public final String responseType;
   public final String clientId;
   public final String sessionId;
+  public final CodeChallenge codeChallenge;
+
+
+  public AuthorizationRequest(String responseType, String clientId, String sessionId, CodeChallenge codeChallenge) {
```
VasilMitovClouway: I think this class is not used anymore.

mgenov: Yes, it's used only in the equality test but it could be used for that we were discussed for the authorization method.

mgenov: You either have to modify this one or deleted this one and introduce a new one. 

Maybe deletion could be performed on separate PR to not mess this one with clearance of the equality test.

VasilMitovClouway: Originally i've added the ClientAuthorizationRequest but in my opinion this name is better. Modified this one and the tests. Now we will be using this for the authorize method of the ClientAuthorizationRepository

oauth2-server/src/main/java/com/clouway/oauth2/AuthCodeAuthorization.java
VasilMitovClouway: Added the new interface. I'm not very sure if this doesn't make things more complex than they should be (for a token to be issues 3 activities are passed) but the dependencies have lowered and the tests are clearer so we have that going for us.

oauth2-server/src/main/java/com/clouway/oauth2/CodeExchangeVerificationFlow.java
```diff
@@ -0,0 +1,43 @@
...
+import com.clouway.oauth2.authorization.Authorization;
+import com.clouway.oauth2.authorization.ClientAuthorizationRepository;
+import com.clouway.oauth2.client.Client;
+import com.clouway.oauth2.codechallenge.CodeVerifier;
+import com.google.common.base.Optional;
+
+/**
+ * @author Vasil Mitov <vasil.mitov@clouway.com>
+ */
+public class CodeExchangeVerificationFlow implements ClientActivity {
```
mgenov: As code exchange requires Authorization for verification it may implement: VerifiedActivity

About the name: VerifiedActivity could become: AuthorizedClientActivity 

```java
Response execute(Authorization authorization, Client client, Request request, DateTime instant);
```

mgenov (changes requested):

oauth2-server/src/main/java/com/clouway/oauth2/CodeExchangeVerificationFlow.java
VasilMitovClouway: I was thinking about that. It becomes a little over the top with naming. We already have AuthorizedClientActivity with construction: 
```java
interface AuthorizedClientActivity {
Response execute(Client client, Identity identity, Set<String> scopes, Request request, DateTime instant, Map<String, String> params);
}
```

Which is responsible for issuing the token.

mgenov: This could become AuthorizedIdentityActivity as identity is passed.

VasilMitovClouway: Another possible way could be IssueTokenActivity/BearerTokenActivity as the result is BearerTokenResponse

mgenov: You always have to return commonly named Response object and it's not explicitly said that BearertokenResponse is returned.

VasilMitovClouway:
PR notes fixed needs another review.

oauth2-server/src/main/java/com/clouway/oauth2/CodeExchangeVerificationFlow.java
VasilMitovClouway: I agree i kept the AuthorizedIdentityActivity

oauth2-server/src/main/java/com/clouway/oauth2/CodeExchangeVerificationFlow.java
```diff
@@ -0,0 +1,43 @@
...
+  public Response execute(Client client, Request request, DateTime instant) {
+    String authCode = request.param("code");
+    String codeVerifierValue = request.param("code_verifier") == null ? "" : request.param("code_verifier");
+    Optional<Authorization> possibleAuthorization = clientAuthorizationRepository.findAuthorization(client, authCode, instant);
+
+    if (!possibleAuthorization.isPresent()) {
+      return OAuthError.invalidGrant();
+    }
+
+    Authorization authorization = possibleAuthorization.get();
```
mgenov: Loading of authorization and verification of code exchange need to be in separate implementations as they are not closely coupled.

// load authorization 
// verify code 
// issue token

VasilMitovClouway: As i finished this change i think we could go one step further and add
//load authorization
//verify code
//load identity
//issue token 
That way every Activity will have 2 dependencies - the action it's doing and for the next activity to be executed. How does that sound ?

VasilMitovClouway: The load identity could be the VerifiedCodeActivity

mgenov: Sounds good.

VasilMitovClouway: Implemented and code has been rearranged

oauth2-server/src/main/java/com/clouway/oauth2/codechallenge/AuthorizationCodeVerifier.java
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ * @author Vasil Mitov <vasil.mitov@clouway.com>
+ */
+public class AuthorizationCodeVerifier implements CodeVerifier {
+  @Override
+  public boolean verify(CodeChallenge codeChallenge, String codeVerifier) {
+    try {
+      //no code verifier was provided and no code challenge was saved so this is a normal OAuth2 code flow
+      if (Strings.isNullOrEmpty(codeVerifier) && isNullOrEmpty(codeChallenge)) {
```
mgenov: codeChallengeCannot be empty as it's not a string value. 

// code challenge was provided or not 
// code challenge was defined or not 
if (!codeChallenge.isProvided()) {
}

but you have to ensure that this instance is always created and null is not passed

oauth2-server/src/main/java/com/clouway/oauth2/codechallenge/AuthorizationCodeVerifier.java
```diff
@@ -0,0 +1,48 @@
...
+import java.nio.charset.StandardCharsets;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+
+/**
+ * @author Vasil Mitov <vasil.mitov@clouway.com>
+ */
+public class AuthorizationCodeVerifier implements CodeVerifier {
+  @Override
+  public boolean verify(CodeChallenge codeChallenge, String codeVerifier) {
```
mgenov: s/codeVerifier/requestedCode/g

oauth2-server/src/main/java/com/clouway/oauth2/CodeExchangeVerificationFlow.java
```diff
@@ -0,0 +1,30 @@
...
+  private final AuthorizedClientActivity authorizedClientActivity;
+
+  CodeExchangeVerificationFlow(CodeVerifier codeVerifier, AuthorizedClientActivity authorizedClientActivity) {
+    this.codeVerifier = codeVerifier;
+    this.authorizedClientActivity = authorizedClientActivity;
+  }
+
+  @Override
+  public Response execute(Authorization authorization, Client client, Request request, DateTime instant) {
+    String codeVerifierValue = request.param("code_verifier") == null ? "" : request.param("code_verifier");
```
mgenov, line 23: Here you ensure that code_verifier is empty string or some value but in the verifier you are checking for null values which smells.

mgenov (changes requested):

oauth2-server/src/main/java/com/clouway/oauth2/CodeExchangeVerificationFlow.java
VasilMitovClouway: I agree that I've over-checked for null. Removed the check in the AuthorizationCodeVerifier.

oauth2-server/src/main/java/com/clouway/oauth2/codechallenge/AuthorizationCodeVerifier.java
VasilMitovClouway: In the ResourceOwnerClientAuthorizationTest it is shown that if it is not provided in the request an empty CodeChallenge is saved.

oauth2-server/src/main/java/com/clouway/oauth2/codechallenge/AuthorizationCodeVerifier.java
VasilMitovClouway: I haven't seen these aliases in a while and i can't really find their meanings. Did this mean replace ?

oauth2-server/src/main/java/com/clouway/oauth2/CodeExchangeVerificationFlow.java
VasilMitovClouway: Also added a test that shows that if it's not provided in the request an empty value will be provided.

oauth2-server/src/main/java/com/clouway/oauth2/codechallenge/AuthorizationCodeVerifier.java
VasilMitovClouway: I don't think requestedCode sounds good. It is providedCode as provided from the client that tries to authenticate. Maby providedCodeVerifier would be better ?

mgenov (approved):
LGTM !

Issue #73, opened 2017-07-12
oauth2/api: pass object instead of list of params


mgenov:
It would be easy for the client apps if an object is passed instead of list of params. If object is passed it may be created with builders for testing and it will simplify changes of this  code. 

For example:
```java
public TokenResponse issueToken(GrantType grantType, Client client, Identity identity, Set<String> scopes, DateTime instant) {
}
```
may be replaced with: 

```java
public TokenResponse issueToken(TokenRequest request) {
  .. 
} 
```
so the test code and production code may use builder-like structures for the creation of the request.

PR #81, opened 2017-12-07
oauth2/api: pass object instead of list of params

VasilMitovClouway:
In Tokens now an object TokenRequest is passed instead of parameters for better readability. 
A builder has been also added.

I'm not sure i should keep these long tests for equality for TokenRequest.

Fixes: #73

### DIFF
--- a/oauth2-server/src/main/java/com/clouway/oauth2/token/TokenRequest.java
+++ b/oauth2-server/src/main/java/com/clouway/oauth2/token/TokenRequest.java
@@ -1,44 +1,100 @@
 package com.clouway.oauth2.token;
 
+import com.clouway.oauth2.DateTime;
+import com.clouway.oauth2.Identity;
+import com.clouway.oauth2.client.Client;
+import com.google.common.base.Objects;
+
+import java.util.Map;
+import java.util.Set;
+
 /**
  * @author Ivan Stefanov <ivan.stefanov@clouway.com>
  */
-public class TokenRequest {
-  public final String grantType;
-  public final String code;
-  public final String refreshToken;
-  public final String clientId;
-  public final String clientSecret;
+public final class TokenRequest {
 
-  public TokenRequest(String grantType, String code, String refreshToken, String clientId, String clientSecret) {
-    this.grantType = grantType;
-    this.code = code;
-    this.refreshToken = refreshToken;
-    this.clientId = clientId;
-    this.clientSecret = clientSecret;
+  public static final class Builder {
+    private GrantType grantType;
+    private Client client;
+    private Identity identity;
+    private Set<String> scopes;
+    private DateTime when;
+    private Map<String, String> params;
+
+    private Builder() {
+    }
+
+    public TokenRequest build() {
+      return new TokenRequest(this);
+    }
+
+    public Builder grantType(GrantType grantType) {
+      this.grantType = grantType;
+      return this;
+    }
+
+    public Builder client(Client client) {
+      this.client = client;
+      return this;
+    }
+
+    public Builder identity(Identity identity) {
+      this.identity = identity;
+      return this;
+    }
+
+    public Builder scopes(Set<String> scopes) {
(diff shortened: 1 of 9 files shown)


oauth2-server/src/main/java/com/clouway/oauth2/token/TokenRequest.java
```diff
@@ -1,44 +1,110 @@
...
-  public final String clientSecret;
+  public final GrantType grantType;
+  public final Client client;
+  public final Identity identity;
+  public final Set<String> scopes;
+  public final DateTime when;
+  public final Map<String, String> params;
 
-  public TokenRequest(String grantType, String code, String refreshToken, String clientId, String clientSecret) {
+  public TokenRequest(GrantType grantType, Client client, Identity identity, Set<String> scopes, DateTime when, Map<String, String> params) {
```
mgenov: This constructor should be removed too and only builder usage to be enforced. 

Passing of many params as method arguments is similar as passing of many params to the constructor.

oauth2-server/src/main/java/com/clouway/oauth2/token/TokenRequest.java
```diff
@@ -1,44 +1,110 @@
...
-    int result = grantType != null ? grantType.hashCode() : 0;
-    result = 31 * result + (code != null ? code.hashCode() : 0);
-    result = 31 * result + (clientId != null ? clientId.hashCode() : 0);
-    result = 31 * result + (clientSecret != null ? clientSecret.hashCode() : 0);
-    return result;
+    return Objects.hashCode(grantType, client, identity, scopes, when, params);
+  }
+
+
+  public static final class Builder {
```
mgenov: Inner classes should be defined first.

oauth2-server/src/test/java/com/clouway/oauth2/token/TokenRequestEqualityTest.java
```diff
@@ -12,17 +20,164 @@
 public class TokenRequestEqualityTest {
   @Test
   public void areEqual() {
-    TokenRequest request1 = new TokenRequest("grant_type", "code","refresh_token", "client_id", "client_secret");
-    TokenRequest request2 = new TokenRequest("grant_type", "code","refresh_token", "client_id", "client_secret");
+    DateTime anyDateTime = new DateTime();
```
mgenov: This test looks a bit overwhelming with few instantiations of whole object. You can simplify it by using 2 cases: equal and not equal. 

EqualsVerifier looks like a library which could be tested about this: http://jqno.nl/equalsverifier/manual/getting-started/

mgenov (changes requested):

oauth2-server/src/test/java/com/clouway/oauth2/token/TokenRequestEqualityTest.java
VasilMitovClouway: With the implementation of this library should i change the Equals tests for other classes as well or should i do it in a separate PR?

oauth2-server/src/main/java/com/clouway/oauth2/token/TokenRequest.java
VasilMitovClouway: Removed.

oauth2-server/src/main/java/com/clouway/oauth2/token/TokenRequest.java
VasilMitovClouway: Moved to be first.

oauth2-server/src/test/java/com/clouway/oauth2/token/TokenRequestEqualityTest.java
mgenov: In separate PR sounds better.

mgenov (approved):
LGTM !

Issue #77, opened 2017-11-27
oauth2: improve error results with description


mgenov:
Error results could be improved with description message in the response. 

Now when token is trying to be refreshed and server fails due some reason is returned only "invalid_grant" but it would give better information if and description is added that better describes the intention.

PR #85, opened 2017-12-22
oauth2: improve error results with description

VasilMitovClouway:
Descriptions have been added to the OAuthErrors occuring in the lbirary so that it will be easier for developers using it to understand what's the problem.

Fixes: #77

### DIFF
--- a/oauth2-server/src/main/java/com/clouway/oauth2/ClientAuthorizationActivity.java
+++ b/oauth2-server/src/main/java/com/clouway/oauth2/ClientAuthorizationActivity.java
@@ -4,8 +4,8 @@ import com.clouway.friendlyserve.Request;
 import com.clouway.friendlyserve.Response;
 import com.clouway.friendlyserve.RsRedirect;
 import com.clouway.oauth2.authorization.Authorization;
-import com.clouway.oauth2.authorization.ClientAuthorizationRepository;
 import com.clouway.oauth2.authorization.AuthorizationRequest;
+import com.clouway.oauth2.authorization.ClientAuthorizationRepository;
 import com.clouway.oauth2.client.Client;
 import com.clouway.oauth2.client.ClientFinder;
 import com.clouway.oauth2.codechallenge.CodeChallenge;
@@ -46,7 +46,7 @@ class ClientAuthorizationActivity implements IdentityActivity {
     Optional<Client> possibleClientResponse = clientFinder.findClient(clientId);
 
     if (!possibleClientResponse.isPresent()) {
-      return OAuthError.unauthorizedClient();
+      return OAuthError.unauthorizedClient(String.format("Unknown client '%s'", clientId));
     }
 
     Client client = possibleClientResponse.get();
--- a/oauth2-server/src/main/java/com/clouway/oauth2/ClientController.java
+++ b/oauth2-server/src/main/java/com/clouway/oauth2/ClientController.java
@@ -25,14 +25,14 @@ class ClientController implements ClientRequest {
 
     // Client was not authorized
     if (!possibleResponse.isPresent()) {
-      return OAuthError.unauthorizedClient();
+      return OAuthError.unauthorizedClient(String.format("Unknown client '%s'", credentials.clientId()));
     }
 
     Client client = possibleResponse.get();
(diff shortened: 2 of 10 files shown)


oauth2-server/src/main/java/com/clouway/oauth2/ClientAuthorizationActivity.java
```diff
@@ -46,7 +46,7 @@ public Response execute(String identityId, Request request) {
     Optional<Client> possibleClientResponse = clientFinder.findClient(clientId);
 
     if (!possibleClientResponse.isPresent()) {
-      return OAuthError.unauthorizedClient();
+      return OAuthError.unauthorizedClient("A client with id:" + clientId + " was either not found or is not authorized.");
```
mgenov: We should be careful with these messages as someone could use them for a brute-force. 

```java
OAuthError.unauthorizedClient(String.format("Unknown client '%s'", clientId))
```

oauth2-server/src/main/java/com/clouway/oauth2/ClientController.java
```diff
@@ -25,14 +25,14 @@ public Response handleAsOf(Request request, ClientCredentials credentials, DateT
 
     // Client was not authorized
     if (!possibleResponse.isPresent()) {
-      return OAuthError.unauthorizedClient();
+      return OAuthError.unauthorizedClient("A client with id:" + credentials.clientId() + " was either not found or is not authorized.");
```
mgenov: OAuthError.unauthorizedClient(String.format("Unknown client '%s'", credentials.clientId()))

oauth2-server/src/main/java/com/clouway/oauth2/ClientController.java
```diff
@@ -25,14 +25,14 @@ public Response handleAsOf(Request request, ClientCredentials credentials, DateT
...
     }
 
     Client client = possibleResponse.get();
 
     // Client credentials did not match?
     if (!client.credentialsMatch(credentials)) {
-      return OAuthError.unauthorizedClient();
+      return OAuthError.unauthorizedClient("Client credentials did not match.");
```
mgenov: The error should be same for unknown and non matching secret as it will not provide information to the caller which to be used for brute force: 

OAuthError.unauthorizedClient(String.format("Unknown client '%s'", clientId))

oauth2-server/src/main/java/com/clouway/oauth2/UserInfoController.java
```diff
@@ -31,14 +31,14 @@ public Response handleAsOf(Request request, DateTime instantTime) {
     Optional<BearerToken> possibleTokenResponse = tokens.findTokenAvailableAt(accessToken, instantTime);
 
     if (!possibleTokenResponse.isPresent()) {
-      return OAuthError.invalidToken();
+      return OAuthError.invalidToken("Access token not found.");
```
mgenov: "Access token was not found."
to match the identity error.

mgenov (changes requested):

oauth2-server/src/main/java/com/clouway/oauth2/ClientAuthorizationActivity.java
VasilMitovClouway: Changed. We may omit the clientId, but that is an information we are already exposing on login.

oauth2-server/src/main/java/com/clouway/oauth2/ClientController.java
VasilMitovClouway: Changed

oauth2-server/src/main/java/com/clouway/oauth2/ClientController.java
VasilMitovClouway: Agreed changed.

oauth2-server/src/main/java/com/clouway/oauth2/UserInfoController.java
VasilMitovClouway: Changed

mgenov (approved):
LGTM !

Issue #84, opened 2017-12-22
example app: auth endpoint called even if the login has failed


nagylzs:
If you try to start an authorization request in the example app and write in a wrong password, then the following happens:

- the LoginEndpoint will redirect to the continue URL (even though the user could not login)
- then the continue url (the auth endpoint) will redirect to the login endpoint again

The biggest "problem" is that the user will not see an error message.

I have a very small modification to the example app that fixes this problem. Can I send a PR?

VasilMitovClouway:
Of course, we would be glad to see the modification.

nagylzs:
Submitted a PR but the tests are failing. I have never used JMock so I'm not sure how to change the tests but I'll try.

mgenov:
You can remove the whole test as it's an example app and it's purpose is to show example usage.

nagylzs:
I feel I totally messed up the PR. I'm sorry I'm new to this. Maybe I should delete this PR and create a new one.

VasilMitovClouway:
No, don't worry.  No one is born taught.

mgenov:
Fixed with #86

PR #86, opened 2017-12-22
example-app: prevent redirect from LoginEndpoint when login fails

nagylzs:
Issue: #84

### DIFF
--- a/oauth2-example-app/src/main/java/com/clouway/oauth2/exampleapp/LoginEndpoint.java
+++ b/oauth2-example-app/src/main/java/com/clouway/oauth2/exampleapp/LoginEndpoint.java
@@ -6,6 +6,9 @@ import com.google.inject.Inject;
 import com.google.sitebricks.Show;
 import com.google.sitebricks.http.Post;
 
+import java.io.UnsupportedEncodingException;
+import java.net.URLEncoder;
+
 import javax.servlet.http.Cookie;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
@@ -17,6 +20,7 @@ import javax.servlet.http.HttpServletResponse;
 public class LoginEndpoint {
   private String username;
   private String password;
+  private String errormessage;
 
   private ResourceOwnerAuthentication authentication;
 
@@ -24,7 +28,7 @@ public class LoginEndpoint {
   public LoginEndpoint(ResourceOwnerAuthentication authentication) {
     this.authentication = authentication;
   }
-
+  
   @Post
   public String login(HttpServletRequest request, HttpServletResponse response) {
     Optional<Session> session = authentication.auth(username, password, request.getRemoteAddr());
@@ -37,9 +41,19 @@ public class LoginEndpoint {
       sid.setPath("/");
 
       response.addCookie(sid);
+      
+      return continueUrl;
+    } else {
+        try {
+			return request.getRequestURI() +					
+					"?continue="+URLEncoder.encode(continueUrl,"UTF-8")+
+				  "&errormessage="+URLEncoder.encode("Invalid username or password","UTF-8");
+		} catch (UnsupportedEncodingException e) {
+			return request.getRequestURI() +
+					"?continue="+URLEncoder.encode(continueUrl)+
+					  "&errormessage="+URLEncoder.encode("Invalid username or password");
+		}
     }
-
-    return continueUrl;
   }
 
   public void setUsername(String username) {
@@ -50,4 +64,11 @@ public class LoginEndpoint {
     this.password = password;
   }
 
+  public String getErrormessage() {
+	return errormessage;
+  }
+
+  public void setErrormessage(String errormessage) {
+	this.errormessage = errormessage;
+  }
 }
--- a/oauth2-example-app/src/test/java/com/clouway/oauth2/exampleapp/LoginEndpointTest.java
+++ b/oauth2-example-app/src/test/java/com/clouway/oauth2/exampleapp/LoginEndpointTest.java
@@ -13,6 +13,8 @@ import javax.servlet.http.HttpServletRequest;
 
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.core.Is.is;
+import static org.junit.Assert.assertTrue;
+
 
 /**
  * @author Ivan Stefanov <ivan.stefanov@clouway.com>
@@ -59,6 +61,7 @@ public class LoginEndpointTest {
       will(Expectations.returnValue(remoteAddress));
       oneOf(request).getParameter("continue");
       will(returnValue("/xxx"));
+      oneOf(request).getRequestURI();
       oneOf(authentication).auth("Grigor", "dimitrov", remoteAddress);
       will(Expectations.returnValue(Optional.absent()));
     }});
@@ -68,7 +71,8 @@ public class LoginEndpointTest {
     String redirectPage = endpoint.login(request, response);
 
     response.assertDoesNotExist("SID");
-    assertThat(redirectPage, is("/xxx"));
+    assertTrue(redirectPage.contains("?continue"));
+    assertTrue(redirectPage.endsWith("&errormessage=Invalid+username+or+password"));
   }
 
   private void pretendThatUserIsAuthorisedWith(String username, String password) {
--- a/oauth2-example-app/src/main/resources/com/clouway/oauth2/exampleapp/login.html
+++ b/oauth2-example-app/src/main/resources/com/clouway/oauth2/exampleapp/login.html
@@ -10,7 +10,7 @@
     <div id="login-title">Example App</div>
 
     <div class="alert-error">
-        <div class="error-message"></div>
+        <div class="error-message">${errormessage}</div>
     </div>
 
     <form method="post">


VasilMitovClouway:
Would be best if you squash your commits to one that will better describe the fix. An example: 
 oauth2/example-app: on login fail a redirect no longer occurs or the current title of the PR is a good candidate as well. We usually try to keep the policy of single commit per PR.

oauth2-example-app/src/main/java/com/clouway/oauth2/exampleapp/LoginEndpoint.java
```diff
@@ -17,6 +20,7 @@
 public class LoginEndpoint {
   private String username;
   private String password;
+  private String errormessage;
```
VasilMitovClouway, line 23: use camelCase when naming - errorMessage

oauth2-example-app/src/main/java/com/clouway/oauth2/exampleapp/LoginEndpoint.java
```diff
@@ -50,4 +57,12 @@ public void setPassword(String password) {
     this.password = password;
   }
 
+  public String getErrormessage() {
```
VasilMitovClouway: These are not used. Usually getters and setters are to be avoided. 
If you want more on the matter : https://dev.to/scottshipp/avoid-getters-and-setters-whenever-possible-c8m

oauth2-example-app/src/main/java/com/clouway/oauth2/exampleapp/storage/InMemoryModule.java
```diff
@@ -36,7 +39,24 @@ protected void configure() {
     bind(SessionSecurity.class).toInstance(resourceOwnerRepository);
 
     InMemoryClientRepository clientRepository = new InMemoryClientRepository();
-    clientRepository.register(new ClientRegistrationRequest("857613db7b18232c72a5093ad19dbc6df74a139e", "test", Collections.singleton("http://localhost:8080/oauth/callback")));
+    
+    String registeredRedirectURI = "https://httpbin.org/get";
```
VasilMitovClouway: Redirect URL looks strange.

oauth2-example-app/src/main/java/com/clouway/oauth2/exampleapp/LoginEndpoint.java
```diff
@@ -26,7 +30,7 @@ public LoginEndpoint(ResourceOwnerAuthentication authentication) {
   }
 
   @Post
-  public String login(HttpServletRequest request, HttpServletResponse response) {
+  public String login(HttpServletRequest request, HttpServletResponse response) throws UnsupportedEncodingException {
```
VasilMitovClouway: Would be better if the error is handled and not thrown by the Post method.

oauth2-example-app/src/main/java/com/clouway/oauth2/exampleapp/storage/InMemoryModule.java
```diff
@@ -36,7 +39,24 @@ protected void configure() {
...
+    Client client = clientRepository.register(new ClientRegistrationRequest("857613db7b18232c72a5093ad19dbc6df74a139e", "test", Collections.singleton(registeredRedirectURI)));
+	try {
+		System.err.println(
```
VasilMitovClouway: We do not have an error here.

oauth2-example-app/src/main/java/com/clouway/oauth2/exampleapp/storage/InMemoryResourceOwnerRepository.java
```diff
@@ -20,6 +20,12 @@
 class InMemoryResourceOwnerRepository implements ResourceOwnerStore, ResourceOwnerAuthentication, SessionSecurity {
   private Map<String, ResourceOwner> resourceOwners = new HashMap<String, ResourceOwner>() {{
     put("admin", new ResourceOwner("admin", "admin"));
+    
+    System.err.println(
```
VasilMitovClouway: .

oauth2-example-app/src/test/java/com/clouway/oauth2/exampleapp/LoginEndpointTest.java
```diff
@@ -13,6 +13,10 @@
 
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.core.Is.is;
+import static org.junit.Assert.assertNotNull;
```
VasilMitovClouway: Unused import.

oauth2-example-app/src/test/java/com/clouway/oauth2/exampleapp/LoginEndpointTest.java
```diff
@@ -13,6 +13,10 @@
...
+import static org.junit.Assert.assertTrue;
+
+import java.net.URLEncoder;
```
VasilMitovClouway: .

VasilMitovClouway (changes requested):

oauth2-example-app/src/main/java/com/clouway/oauth2/exampleapp/LoginEndpoint.java
```diff
@@ -50,4 +57,12 @@ public void setPassword(String password) {
     this.password = password;
   }
 
+  public String getErrormessage() {
+	return errormessage;
+  }
+
+  public void setErrormessage(String errormessage) {
```
VasilMitovClouway, line 71: You are not posting the message from the login page. The username and password setters are used by sitebricks to set them but the error message is not posted so it is never set or get.

VasilMitovClouway (changes requested):

oauth2-example-app/src/main/java/com/clouway/oauth2/exampleapp/LoginEndpoint.java
nagylzs: Okay should I throw an Error instead like this?

 catch(... e) {throw new Error("The world is ending, UTF-8 is not an encoding anymore!", e);}

or maybe I could use URLEncoder.encode without the encoding parameter (but that is deprecated API)

oauth2-example-app/src/main/java/com/clouway/oauth2/exampleapp/storage/InMemoryModule.java
nagylzs: It is just for the example, it redirects to a page that displays all parameters on a page, so anyone can try example oauth server alone (without creating another server to be redirected to), without modifications.

oauth2-example-app/src/main/java/com/clouway/oauth2/exampleapp/LoginEndpoint.java
nagylzs: It it not camelCase because the LoginEndpoint receives it as a get parameter. If I change it to errorMessage then the GET parameter must also be camelCase.

oauth2-example-app/src/main/java/com/clouway/oauth2/exampleapp/LoginEndpoint.java
nagylzs: The LoginPage redirects to itself with &errormessage=Invalid+username+or+password and that sets the field.

nagylzs:
Removed the unnecessary changes, removed unused imports, fixed the LoginEndpointTest. The tests are passing. The only change I did not make was the errorMessage camelcase and the setter, because it is needed to display the error message. I'll try to do everything in a single commit next time.

oauth2-example-app/src/test/java/com/clouway/oauth2/exampleapp/LoginEndpointTest.java
```diff
@@ -59,6 +61,7 @@ public void wrongAuthentication() throws Exception {
       will(Expectations.returnValue(remoteAddress));
       oneOf(request).getParameter("continue");
       will(returnValue("/xxx"));
+      allowing(request).getRequestURI();
```
VasilMitovClouway: It's a bad idea to use allowing. This means that it can be called any number of times, and if you write 100 lines of request.getRequestURI the test will still pass. In the future if you wish to show that a method is called many times it would be better to specify how many times it is called with *exactly(n).of*. 
Here's a cheat sheet for jMock http://www.jmock.org/cheat-sheet.html

VasilMitovClouway (changes requested):

oauth2-example-app/src/main/java/com/clouway/oauth2/exampleapp/storage/InMemoryResourceOwnerRepository.java
VasilMitovClouway: I think this message is not needed.

nagylzs: Message removed.

oauth2-example-app/src/test/java/com/clouway/oauth2/exampleapp/LoginEndpointTest.java
nagylzs: Changed to OneOf

oauth2-example-app/src/main/java/com/clouway/oauth2/exampleapp/storage/InMemoryModule.java
nagylzs: Changes removed

oauth2-example-app/src/main/java/com/clouway/oauth2/exampleapp/storage/InMemoryModule.java
nagylzs: Changes removed, anyway

VasilMitovClouway:
We will wait for @mgenov to take a look at the PR as well. While we wait you could do from the console 
git rebase -i HEAD~[number of commits] to rebase the last  commits into one

mgenov (approved):
LGTM !

You have to squash your commits and do force update on your origin to be able to merge it. 

git remote add upstream git@github.com:clouway/oauth2-server.git
git fetch upstream
git rebase upstream/master -i 
git push origin master -f

nagylzs:
Squashed commits into one, force pushed

PR #88, opened 2018-03-07
oauth2-server: OAuth client id is added to authorization params

qnislavnachev:
OAuth client id is passed to the authorization params,
so that way in any state of the oauth authorization,
client applications will be able to access client info.

### DIFF
--- a/oauth2-server/src/main/java/com/clouway/oauth2/ClientAuthorizationActivity.java
+++ b/oauth2-server/src/main/java/com/clouway/oauth2/ClientAuthorizationActivity.java
@@ -41,7 +41,7 @@ class ClientAuthorizationActivity implements IdentityActivity {
     String codeVerifierMethod = request.param("code_challenge_method") == null ? "" : request.param("code_challenge_method");
     CodeChallenge codeChallenge = new CodeChallenge(codeChallengeValue, codeVerifierMethod);
 
-    Map<String, String> params = new Params().parse(request, "response_type", "client_id", "redirect_uri", "state", "scope", "code_challenge", "code_challenge_method");
+    Map<String, String> params = new Params().parse(request, "response_type", "redirect_uri", "state", "scope", "code_challenge", "code_challenge_method");
 
     Optional<Client> possibleClientResponse = clientFinder.findClient(clientId);
 
--- a/oauth2-server/src/main/java/com/clouway/oauth2/ClientAuthorizationActivity.java
+++ b/oauth2-server/src/main/java/com/clouway/oauth2/ClientAuthorizationActivity.java
@@ -41,7 +41,7 @@ class ClientAuthorizationActivity implements IdentityActivity {
     String codeVerifierMethod = request.param("code_challenge_method") == null ? "" : request.param("code_challenge_method");
     CodeChallenge codeChallenge = new CodeChallenge(codeChallengeValue, codeVerifierMethod);
 
-    Map<String, String> params = new Params().parse(request, "response_type", "redirect_uri", "state", "scope", "code_challenge", "code_challenge_method");
+    Map<String, String> excludedParams = new Params().parse(request, "response_type", "client_id", "redirect_uri", "state", "scope", "code_challenge", "code_challenge_method");
 
     Optional<Client> possibleClientResponse = clientFinder.findClient(clientId);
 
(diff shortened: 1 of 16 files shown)


mgenov:
The description says that clientId was added, but this is what I see in this PR: 

screen shot 2018-03-07 at 14 35 54

qnislavnachev:
@mgenov This strings right there are "excludes". This means which request params should be excluded from the custom params. In that case I remove "client_id" so in the custom params client id will be added.

mgenov:
So naming could be improved to excludedParams ?

qnislavnachev:
Good point. I am gonna rename it.

oauth2-server/src/main/java/com/clouway/oauth2/ClientAuthorizationActivity.java
```diff
@@ -41,7 +41,7 @@ public Response execute(String identityId, Request request) {
     String codeVerifierMethod = request.param("code_challenge_method") == null ? "" : request.param("code_challenge_method");
     CodeChallenge codeChallenge = new CodeChallenge(codeChallengeValue, codeVerifierMethod);
 
-    Map<String, String> params = new Params().parse(request, "response_type", "client_id", "redirect_uri", "state", "scope", "code_challenge", "code_challenge_method");
+    Map<String, String> params = new Params().parse(request, "response_type", "redirect_uri", "state", "scope", "code_challenge", "code_challenge_method");
```
mlesikov: I can agree with this change, but I'm not sure that this is the best solution. 
The other solution could 
```
Optional<Identity> possibleIdentity = identityFinder.findIdentity(authorization.identityId, GrantType.AUTHORIZATION_CODE, instant, authorization.params, **authorization.clientId**);
```
adding the clientId to the identityFinder.findIdentity(...) method

the main purpose is the clientId to be provided to the identityFinder.findIdentity(...), so we can fetch the correct customer in case that the client is not internal and is owned from a single customer.

qnislavnachev: I am agree with your suggestion, but doesnt it look strange for the method 
``` findIdentity(.. clientId: String) ``` to pass oauth client id that is used for authorization, when you are looking for identity ?

Comment (mgenov) [reply oauth2-server/src/main/java/com/clouway/oauth2/ClientAuthorizationActivity.java]: If this change adds clientId to the authorization object it sounds reasonable to me.

Comment (qnislavnachev) [reply oauth2-server/src/main/java/com/clouway/oauth2/ClientAuthorizationActivity.java]: No this change doesnt add clientId to the authorization object, the authorization object already has client id as a field

Comment (mgenov) [reply oauth2-server/src/main/java/com/clouway/oauth2/ClientAuthorizationActivity.java]: This answer doesn't help me to figure this out. 

This is what I see in tests: 
```java
oneOf(clientAuthorizationRepository).authorize(new AuthorizationRequest(anyExistingClient, "user1", "code", Collections.singleton("abc"), emptyCodeChallenge, ImmutableMap.of("customerIndex", "::customerIndex::", "client_id", "::client_id::")));
will(returnValue(Optional.of(new Authorization("code", "::client_id::", "identityId", "1234", Collections.<String>emptySet(), Collections.singleton("http://example.com/callback"), Collections.singletonMap("customer", "::customerIndex::")))));
```
e.g 

the clientAuthorizationRepository.authorize is called with clientID and returns Authorization object with additional custom params. Note that this "customerId" is moved back to the library which doesn't sound reasonable as it couples it with internal of an app.

qnislavnachev: This PR is for that to provide clientId  to identity finder. With that client the finder will be able to select the "customer" of users. This is the purpose of this change.

mgenov: You mean that it will return an identity with appropriate custom claims associated with it ?

qnislavnachev:
Added second commit where client id is passed directly to the FindAuthorizationRequest object. No longer client id is passed in excluded params

oauth2-server/src/main/java/com/clouway/oauth2/ClientAuthorizationActivity.java
qnislavnachev: Yes

mgenov (approved):
LGTM !

PR #90, opened 2018-05-08
oauth2-server: on refresh token no longer auth header is required

qnislavnachev:
authorization header was required when refreshing token,
that check was removed


mgenov (approved):
LGTM !

PR #92, opened 2018-10-10
oauth2-server: code in authorization reseponse is url encoded

mgenov:
The code in the authorization response url is now url encoded and allows not url safe strings to be returned as code.

Thanks to Ivan Surzhenko for the feedback.


tsbonev (approved):
LGTM !

PR #93, opened 2018-10-19
oauth2/integration: better library integration

mlesikov:
The library had a strog dependency to the HttpServlet. All the endpoint
setups were configured in the OAuth2Servlet, so when we try to integrate
the oauth2 library in our code we should use this servlet. This makes
the integration a litle bit restrictive and cause troubles if would like
to use frameworks like SparkJava.
Now the configurations logic was moved out and separated from the need to
use directly the HttpServlet in your project

SparkJava integration example : 
```kotlin
fun main(args: Array<String>) {

    val tokens = null
    val jwtKeyStore = null
    val keyStore = null
    val identityFinder = null
    val resourceOwnerIdentityFinder = null
    val clientAuthorizationRepository = null
    val clientFinder = null
    val oauth2Config  = OAuth2Config.newConfig()
                .tokens(tokens)
                .jwtKeyStore(jwtKeyStore)
                .keyStore(keyStore)
                .identityFinder(identityFinder)
                .resourceOwnerIdentityFinder(resourceOwnerIdentityFinder)
                .clientAuthorizationRepository(clientAuthorizationRepository)
                .clientFinder(clientFinder)
                .loginPageUrl("/ServiceLogin?continue=")
                .build()
    val oauth2ServletApi = Oauth2ServlertApiSupportFactory().create(oauth2Config)

    Spark.path("/o/oauth2/v1/") {
        Spark.get("*", Oauth2Handler(oauth2ServletApi))
        Spark.post("*", Oauth2Handler(oauth2ServletApi))
    }

    println("App is up and running")
}

class Oauth2Handler(private val oauth2ServletApi: Oauth2ServlertApiSupport) : Route {

    override fun handle(request: Request?, response: Response?): Any? {

        oauth2ServletApi.serve(request!!.raw(), response!!.raw())

        return response.raw()
    }
}
```

### DIFF
--- a/oauth2-server/src/main/java/com/clouway/oauth2/OAuth2ApiSupport.java
+++ b/oauth2-server/src/main/java/com/clouway/oauth2/OAuth2ApiSupport.java
@@ -1,0 +1,14 @@
+package com.clouway.oauth2;
+
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import java.io.IOException;
+
+/**
+ * @author Mihail Lesikov (mihail.lesikov@clouway.com)
+ */
+public interface OAuth2ApiSupport {
+
+  void serve(HttpServletRequest req, HttpServletResponse resp) throws IOException;
+
+}
--- a/oauth2-server/src/main/java/com/clouway/oauth2/OAuth2ApiSupportFactory.java
+++ b/oauth2-server/src/main/java/com/clouway/oauth2/OAuth2ApiSupportFactory.java
@@ -1,0 +1,131 @@
+package com.clouway.oauth2;
+
+import com.clouway.friendlyserve.FkRegex;
+import com.clouway.friendlyserve.RequestHandlerMatchingParam;
+import com.clouway.friendlyserve.RequiresParam;
+import com.clouway.friendlyserve.TkFork;
+import com.clouway.friendlyserve.servlets.ServletApiSupport;
+import com.clouway.oauth2.codechallenge.AuthorizationCodeVerifier;
+import com.clouway.oauth2.jws.RsaJwsSignature;
+import com.clouway.oauth2.jws.Signature;
+import com.clouway.oauth2.jws.SignatureFactory;
+import com.clouway.oauth2.jwt.Jwt.Header;
+import com.clouway.oauth2.jwt.JwtController;
+import com.clouway.oauth2.token.JjwtIdTokenFactory;
+import com.google.common.base.Optional;
+
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import java.io.IOException;
+
+/**
+ * @author Mihail Lesikov (mihail.lesikov@clouway.com)
+ */
+public class OAuth2ApiSupportFactory {
+
+  public OAuth2ApiSupport create(OAuth2Config config) {
+    final SignatureFactory signatureFactory = new SignatureFactory() {
+      @Override
+      public Optional<Signature> createSignature(byte[] signatureValue, Header header) {
+        // It's vulnerable multiple algorithms to be supported, so server need to reject
+        // any calls from not supported algorithms
+        // More infromation could be taken from: https://auth0.com/blog/2015/03/31/critical-vulnerabilities-in-json-web-token-libraries/
+        if (!"RS256".equals(header.alg)) {
+          return Optional.absent();
+        }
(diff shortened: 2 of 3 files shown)


oauth2-server/src/main/java/com/clouway/oauth2/Oauth2ServlertApiSupportFactory.java
```diff
@@ -0,0 +1,131 @@
...
+import com.google.common.base.Optional;
+
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import java.io.IOException;
+
+/**
+ * @author Mihail Lesikov (mihail.lesikov@clouway.com)
+ */
+public class Oauth2ServlertApiSupportFactory {
```
mgenov: typo: OAuth2Servlet

oauth2-server/src/main/java/com/clouway/oauth2/Oauth2ServlertApiSupport.java
```diff
@@ -0,0 +1,14 @@
+package com.clouway.oauth2;
+
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import java.io.IOException;
+
+/**
+ * @author Mihail Lesikov (mihail.lesikov@clouway.com)
+ */
+public interface Oauth2ServlertApiSupport {
```
mgenov: typo: OAuth2ServletApiSupport

mgenov (changes requested):

oauth2-server/src/main/java/com/clouway/oauth2/Oauth2ServlertApiSupportFactory.java
mlesikov: fixed

oauth2-server/src/main/java/com/clouway/oauth2/Oauth2ServlertApiSupport.java
mlesikov: fixed

oauth2-server/pom.xml
```diff
@@ -4,11 +4,11 @@
   <parent>
     <groupId>com.clouway.security</groupId>
     <artifactId>parent</artifactId>
-    <version>1.0.6-SNAPSHOT</version>
+    <version>1.0.7-SNAPSHOT</version>
```
mgenov: Dumping happens automatically and this change is not required.

mlesikov: I've changed and the version of the library

mgenov: http://maven.apache.org/maven-release/maven-release-plugin/

mlesikov: ok I'll roll it back

oauth2-server/src/main/java/com/clouway/oauth2/OAuth2ServletApiSupport.java
```diff
@@ -0,0 +1,14 @@
+package com.clouway.oauth2;
+
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import java.io.IOException;
+
+/**
+ * @author Mihail Lesikov (mihail.lesikov@clouway.com)
+ */
+public interface OAuth2ServletApiSupport {
```
VasilMitovClouway: Maby just OAuth2ApiSupport so that Servlet is not involved.

mlesikov: sounds good to me

mgenov: It's related to Servlet API as method accepts HttpServletRequest and HttpServletResponse.

A better definition of this interface need to be provided which to not be related to the Servlet API. If it accepts friendly serve requests and responses then the client app which is using the library will be able to create adapter depending on the environment.

mgenov: It could happen in following PR btw.

mgenov (approved):
LGTM !

mgenov:
1.0.6 is now released and snapshot version of 1.0.7 is now available in sonatype: https://oss.sonatype.org/content/repositories/snapshots/com/clouway/security/oauth2-server/1.0.7-SNAPSHOT/

In an hour it will be available and in maven central.

PR #95, opened 2018-12-16
oauth2-example-app: update version of jetty

mgenov:
The Jetty version in the example app is deployed on jetty.


mgenov:
retest this please

